Issue #8, opened 2015-04-10
Copy-edit all existing content for Markdown compliance


RyanCavanaugh:
I did a rough transition from CodePlex wiki to Markdown, but we need to do a detailed scan of each page to look for any formatting problems.

- [x] Two spaces after periods
- [x] Language tags after triple-ticks (need a bunch of these)
- [ ] Headings level (standardize on one-hash as top-level heading size)
- [ ] Fix link formatting

RyanCavanaugh:
Things to watch out for:
 * Two spaces after periods
 * Language tags after triple-ticks (need a bunch of these)
 * Headings level (standardize on one-hash as top-level heading size)

Edit this comment with more as you see them

DanielRosenwasser:
I've turned this issue into a task list.

mhegazy:
should be fixed now.

PR #25, opened 2015-04-24
Remove excess spaces

DanielRosenwasser:
Fixes the excess space formatting issue mentioned in #8.

### DIFF
--- a/pages/Generics.md
+++ b/pages/Generics.md
@@ -6,7 +6,7 @@ In languages like C# and Java, one of the main tools in the toolbox for creating
 
 # Hello World of Generics
 
-To start off, let's do the "hello world" of generics: the identity function. The identity function is a function that will return back whatever is passed in. You can think of this in a similar way to the 'echo' command. 
+To start off, let's do the "hello world" of generics: the identity function. The identity function is a function that will return back whatever is passed in. You can think of this in a similar way to the 'echo' command.
 
 Without generics, we would either have to give the identity function a specific type:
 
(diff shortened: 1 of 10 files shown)


mhegazy:
:+1:

pages/Type Inference.md
```diff
@@ -1,66 +1,66 @@
...
 ```
-window.onmousedown = function(mouseEvent) { 
-    console.log(mouseEvent.buton);  //<- Error  
+window.onmousedown = function(mouseEvent) {
+    console.log(mouseEvent.buton);  //<- Error
 };
 ```
 
-For the code above to give the type error, the TypeScript type checker used the type of the Window.onmousedown function to infer the type of the function expression on the right hand side of the assignment.  When it did so, it was able to infer the type of the `mouseeVent` parameter.  If this function expression were not in a contextually typed position, the `mouseeVent` parameter would have type `any`, and no error would have been issued.
+For the code above to give the type error, the TypeScript type checker used the type of the Window.onmousedown function to infer the type of the function expression on the right hand side of the assignment. When it did so, it was able to infer the type of the `mouseeVent` parameter. If this function expression were not in a contextually typed position, the `mouseeVent` parameter would have type `any`, and no error would have been issued.
```
DanielRosenwasser: `mouseeVent`

pages/Generics.md
```diff
@@ -147,7 +147,7 @@ function identity<T>(arg: T): T {
 var myIdentity: GenericIdentityFn = identity;
 ```
 
-In a similar example, we may want to move the generic parameter to be a parameter of the whole interface. This lets us see what type(s) we're generic over (eg Dictionary<string> rather than just Dictionary). This makes the type parameter visible to all the other members of the interface. 
+In a similar example, we may want to move the generic parameter to be a parameter of the whole interface. This lets us see what type(s) we're generic over (eg Dictionary<string> rather than just Dictionary). This makes the type parameter visible to all the other members of the interface.
```
DanielRosenwasser: eg -> e.g., use code style for Dictionary/Dictionary<string>

PR #30, opened 2015-04-29
Fix typo: x and y are inversed

zhongsp:
Please review whether `x` and `y` are inversed or not here.


DanielRosenwasser:
Reviewed with @JsonFreeman. Wording is a little weird here though. We should review it.

Thanks!

PR #41, opened 2015-04-30
Add a '.gitattributes' file and normalize all line endings.

DanielRosenwasser:
Since we don't have any of the same issues with baselines, let's just do the right thing here.


mhegazy:
:+1:

PR #47, opened 2015-05-23
Use GitBook to publish the handbook

DanielRosenwasser:
This PR makes appropriate changes so that we may publish the handbook through [GitBook](https://www.gitbook.com), a utility to package Markdown files into several formats - most importantly, HTML.

Pros:
* GitBook runs on Node and can be acquired via npm
* GitBook is free, open source, and Apache 2.0
* The default-generated site is really nice. It comes with
    * Search capabilities.
    * Different font style/size selection
    * Different reading modes (e.g. sepia, night mode)
    * Table of contents
    * Tracked sections for each table of contents.
* Simple enough that I learned about it a few hours ago and managed to get it set up pretty quickly.

Cons:
* The default-generated site says "Published with GitBook", but really not a big deal.
* The syntax highlighting is a little busted.

# Try it out

```sh
npm install -g gitbook-cli
gitbook serve pages
```

Now open a browser to `localhost:4000`.

# Screenshots!

Let's be fair, this is really what you came to see.

image

image

image

image

# Acknowledgements

Who else but @basarat with his [typescript-book](https://github.com/basarat/typescript-book) repository?


basarat:
nice :) :rose:

DanielRosenwasser:
Note that Gitbook 2.0.2 seems to be broken with this - it can be amended by running something like the following before serving.

```sh
gitbook versions:install 2.0.1
gitbook versions:uninstall 2.0.2
```

mhegazy:
What about customization? can we change the style sheets? headers, add backgrounds.. etc.. i am thinking of hosting it in typescriptlang.org, what do we need to do to get that looking like the rest of the site?

DanielRosenwasser:
There seem to be plugins we could use for this; probably depends on how easily we can integrate existing stylesheets.

danquirk:
Could you elaborate on what busted syntax highlighting means? That is an important feature for our docs :)

PR #49, opened 2015-05-27
Fix typo: syntax error in template strings

zhongsp:


DanielRosenwasser:
My mistake - thanks for catching this @zhongsp!

PR #50, opened 2015-05-30
Move sentences to different lines for easier diffing

DanielRosenwasser:
Probably should just view the rich diff and glance at the regular diff.


mhegazy:
:+1:

Issue #36, opened 2015-04-29
Rename internal modules to namespaces


mhegazy:

DanielRosenwasser:
Can I take this on? Do we want to mention the "module" keyword apart from when declaring an external module?

danquirk:
For now I think we should keep the internal modules topic and just have one line explaining the rename and link to the new namespaces section. Likewise the namespace section should call out clearly (not some tucked away parenthetical) that namespaces were previously called internal modules. So many existing blog posts, StackOverflow answers, etc refer to internal modules we want to make sure it's easy/clear to someone coming to our docs afterwards what has happened.

DanielRosenwasser:
Fixed by #51.

PR #51, opened 2015-06-02
Update mentions of internal/external modules to use namespaces/modules

DanielRosenwasser:
Addresses #36. Not sure if this is what you had in mind @danquirk but I'd definitely appreciate your feedback.

As usual, rich diff is best diff.


pages/Functions.md
```diff
@@ -2,7 +2,7 @@
 
 Functions are the fundamental building block of any applications in JavaScript.
 They're how you build up layers of abstraction, mimicking classes, information hiding, and modules.
-In TypeScript, while there are classes and modules, function still play the key role in describing how to `do` things.
+In TypeScript, while there are classes, namespaces, and modules, function still play the key role in describing how to *do* things.
```
DanielRosenwasser: This emphasis seems silly, but I'm willing to keep it.

pages/Namespaces and Modules.md
```diff
@@ -393,18 +409,18 @@ In some cases, you may want to only load a module under some conditions.
 In TypeScript, we can use the pattern shown below to implement this and other advanced loading scenarios to directly invoke the module loaders without losing type safety.
 
 The compiler detects whether each module is used in the emitted JavaScript.
-For modules that are only used as part of the type system, no require calls are emitted.
+For modules that are only used as part of the type system, no `require` calls are emitted.
```
danquirk: This wording is a little odd ('part of the type system'). I would say like "If a module identifier is only ever used in type annotations and never as an expression then no 'require' call is emitted for that module."

pages/Namespaces and Modules.md
```diff
@@ -499,27 +519,26 @@ import url = require("url");
...
 
+# Pitfalls of Namespaces and Modules
 
-# Pitfalls of Modules
-
-In this section we'll describe various common pitfalls in using internal and external modules, and how to avoid them.
+In this section we'll describe various common pitfalls in using namespaces and modules, and how to avoid them.
 
-## `/// <reference>`-ing an external module
+## `/// <reference>`-ing a module
```
danquirk: I know this isn't related to your change but it sure is awkward to list this pitfall right after an example where node.d.ts has to mix a /// and an import

danquirk:
:+1:

pages/Namespaces and Modules.md
DanielRosenwasser: Agreed, let's fix this up in a different change though.

PR #55, opened 2015-07-29
Missing 'we' word in the 'Generics' doc

Gillespie59:


DanielRosenwasser:
Thanks! I also notice that should be a question mark. I'll take care of it.

Issue #35, opened 2015-04-29
Union types


mhegazy:
- [x] Union types
- [x] type guards

PR #74, opened 2015-10-13
Cover union types, type aliases, type guards

### LINKED ISSUE #37: type aliases
Possibly lumped with interfaces.

DanielRosenwasser:
Fixes #37 and partially addresses #35

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -43,8 +43,8 @@ let indentedString = padLeft("Hello world", true); // errors during compilation
 A union type describes a value that can be one of several types.
 We use the vertical bar (`|`) to separate each type, so `number | string | boolean` is the type of a value that can be a `number`, a `string`, or a `boolean`.
 
-If we have a value that has a union type, we can only access members that are common to all types in the union. 
- 
+If we have a value that has a union type, we can only access members that are common to all types in the union.
+
 ```ts
 interface Bird {
     fly();
@@ -69,7 +69,7 @@ Union types can be a bit tricky here, but it just takes a bit of intuition to ge
 If a value has the type `A | B`, we only know for *certain* that it has members that both `A` *and* `B` have.
 On the other hand, in the example above `Bird` has a member named `fly`.
 We can't be sure whether a variable typed as `Bird | Fish` has a `fly` method.
-If the variable is really a `B` at runtime, then calling `pet.fly()` will fail.  
+If the variable is really a `B` at runtime, then calling `pet.fly()` will fail.
 
 # Type Guards and Differentiating Types
 
@@ -118,7 +118,7 @@ function isFish(pet: Fish | Bird): pet is Fish {
 }
 ```
 
-Any time `isFish` is used on a variable name, TypeScript will know that variable has that specific type. 
+Any time `isFish` is used on a variable name, TypeScript will know that variable has that specific type.
 
 ```ts
 // Both calls to 'swim' and 'fly' are now okay.
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -25,7 +25,7 @@ That means that we can call it with an argument that's neither a `number` nor a 
 let indentedString = padLeft("Hello world", true); // passes at compile time, fails at runtime.
 ```
 
-Instead of `any`, we can use a *union type* for `padding`:
+Instead of `any`, we can use a *union type* for the `padding` parameter:
 
 ```TypeScript
 /**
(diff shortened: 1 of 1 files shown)


pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
+# Introduction
+
+TODO
```
weswigham: Let me write one for you if that's okay:

> When thinking about types, sometimes it is useful to think about operations on those types -- if we have something which is both a `Fridge` and a `Toaster`, how can we best describe this amalgamation?
> Well, usually in the same way we just did -- as a `Fridge & Toaster`.
> This kind of basic type arithmetic can be incredibly useful in being more expressive with your types, helping you say what you mean without jumping through classical inheritance-based hoops.
> In TypeScript, we support three main productions which operate on types - [Unions](# Union Types), [Intersections](# Intersection Types), and [Aliases](# Type Aliases).
> Supporting these we have [Type Guards](# Type Guards), boolean-returning functions meant to indicate to the typesystem weather an argument is to be considered of a narrowed type in a given branch of code.
> Using these constructs correctly can enhance the semantic meaning of your types and help create a strongly typed, meaningful interface for your TypeScript program.

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+# Type Aliases
+
+Type aliases create a new name for a type.
+Type aliases are sometimes similar to interfaces, but can name primitives, unions, tuples, and any other types that you'd otherwise have to write by hand.
+
+```TypeScript
+type XCoord = number;
+type YCoord = number;
+
+type XCoord = { x: XCoord };
```
weswigham: Duplicate identifier 'XCoord'.

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+let pet = getSmallPet();
+pet.layEggs(); // okay
+pet.swim();    // errors
+```
+
+Union types can be a bit tricky here, but it just takes a bit of intuition to get used to.
+If a value has the type `A | B`, we only know for *certain* that it has members that both `A` *and* `B` have.
+On the other hand, in the example above `Bird` has a member named `fly`.
+We can't be sure whether a variable typed as `Bird | Fish` has a `fly` method.
+If the variable is really a `B` at runtime, then calling `pet.fly()` will fail.
```
sandersn: If it's really a `Fish`.

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+}
+
+let pet = getSmallPet();
+pet.layEggs(); // okay
+pet.swim();    // errors
+```
+
+Union types can be a bit tricky here, but it just takes a bit of intuition to get used to.
+If a value has the type `A | B`, we only know for *certain* that it has members that both `A` *and* `B` have.
+On the other hand, in the example above `Bird` has a member named `fly`.
```
sandersn: I would reword as "In this example, `Bird` ..."

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+    (<Bird>pet).fly();
+}
+```
+
+## User Defined Type Guards
+
+Notice that we had to use type assertions several times.
+It would be much better if once we performed the check, we could know the type of `pet` within each branch.
+
+It just so happens that TypeScript has something called a *type guard*.
```
sandersn: I would say something more like "To support this, Typescript has *type guards*.

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+A type guard is some expression that performs a runtime check that guarantees the type in some scope.
```
sandersn: "**an** expression"

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+We can write a type guard using a regular function:
```
sandersn: I would just say "a function".

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+    return (<Fish>pet).swim !== undefined;
+}
+```
+
+Any time `isFish` is used on a variable name, TypeScript will know that variable has that specific type. 
+
+```TypeScript
+// Both calls to 'swim' and 'fly' are now okay.
+
+if (isFish(pet) {
```
sandersn: missing close parenthesis: `if (isFish(pet))

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+TODO
+
+## `instanceof` type guards
+
+TODO
+
+# Type Aliases
+
+Type aliases create a new name for a type.
+Type aliases are sometimes similar to interfaces, but can name primitives, unions, tuples, and any other types that you'd otherwise have to write by hand.
```
sandersn: A *little* more detail would be nice here -- my first thought was, "Why not get rid of interfaces then?" Specifically, I'd like a short explanation of what aliases can't do that interfaces can.

pages/Advanced Types.md
```diff
@@ -0,0 +1,197 @@
...
+
+type Coordinate = XCoord | XYCoord | XYZCoord;
+type CoordList = Coordinate[];
+
+let coord: CoordList = [{ x: 10, y: 10}, { x: 0, y: 42, z: 10 }, { x: 5 }];
+```
+
+Aliasing doesn't actually create a new type - it creates a new *name* to refer to that type.
+So `10` is a perfectly valid `XCoord` and `YCoord` because they both just refer to `number`.
+Aliasing a primitive is not terribly useful, though it can be used as a form of documentation.
```
sandersn: Specifically, it sounds like aliases don't do the Haskell `newtype` thing where you can't assign between instances of the alias types:

```ts
var x: XCoord = 10;
var y: YCoord = x; // probably NO error
var z: XCoord = y; // probably NO error
```

It might be worthwhile to note that.

weswigham: I think that's what `*name*` was for.

pages/Advanced Types.md
DanielRosenwasser: That's pretty great, though, I'm not 100% certain we're going to have intersection types here; it would be appropriate though.

DanielRosenwasser:
@sandersn and @weswigham if you have feedback about recent changes, I'd be open to still making amendments.

sandersn:
I still don't like "It just so happens Typescript has type guards". :persevere: 
That's the only thing -- the more recent changes look good.

PR #78, opened 2015-10-13
Decorators

rbuckton:
Adds a topic for Decorators.

### DIFF
--- a/pages/Decorators.md
+++ b/pages/Decorators.md
@@ -1,12 +1,12 @@
 # Introduction
 
 With the introduction of Classes in TypeScript and ES6, there now exist certain scenarios that require additional features to support annotating or modifying classes and class members.
-*Decorators* provide a way to add both annotations and a meta-programming syntax for class declarations and members.
-*Decorators* are a [stage 1 proposal](https://github.com/wycats/javascript-decorators/blob/master/README.md) for JavaScript and are available as an experimental feature of TypeScript.
+Decorators provide a way to add both annotations and a meta-programming syntax for class declarations and members.
+Decorators are a [stage 1 proposal](https://github.com/wycats/javascript-decorators/blob/master/README.md) for JavaScript and are available as an experimental feature of TypeScript.
 
 > NOTE&emsp; Decorators are an experimental feature that may change in future releases.
 
-To enable experimental support for *Decorators*, you must enable the `experimentalDecorators` compiler option either on the command line or in your `tsconfig.json`:
+To enable experimental support for decorators, you must enable the `experimentalDecorators` compiler option either on the command line or in your `tsconfig.json`:
 
 **Command Line**:
 ```
@@ -40,8 +40,8 @@ function sealed(target) {
 
 ## Decorator Factories
 
-If we want to customize how a decorator is applied to a declaration, we can write a *Decorator Factory*.
-A *Decorator Factory* is simply a function that returns the expression that will be called at runtime.
+If we want to customize how a decorator is applied to a declaration, we can write a decorator factory.
+A *Decorator Factory* is simply a function that returns the expression that will be called by the decorator at runtime.
 
 We can write a decorator factory in the following fashion:
 
@@ -80,7 +80,7 @@ As such, the following steps are performed when evaluating multiple decorators o
 1. The expressions for each decorator are evaluated top-to-bottom.
 2. The results are then called as functions from bottom-to-top.
 
-If we were to use decorator factories, we can observe this evaluation order with the following example:
+If we were to use [decorator factories](#decorator-factories), we can observe this evaluation order with the following example:
 
 ```ts
 function f() {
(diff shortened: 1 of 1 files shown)


pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
...
+
+# Decorators
+
+A *Decorator* is a special kind of declaration that can be attached to a [class declaration](#class-decorators), [method](#method-decorators), [accessor](#accessor-decorators), [property](#property-decorators), or [parameter](#parameter-decorators).
+Decorators use the form `@expression`, where `expression` must evaluate to a function that will be called at runtime with information about the decorated declaration.
+
+For example, given the decorator `@sealed` we might write the `sealed` function as follows:
+
+```ts
+function sealed(target) {
```
sandersn: does Typescript infer the type of target here? or it is `any`? I notice that it's not annotated in the other examples I've read, so maybe it's worth calling out if people will be forced to type it as `any`.

rbuckton: It's just `any` here, but I wasn't trying to get into anything deeply complex about decorators at this point, just illustrating that the result of the expression must be \[\[Call]]able.

sandersn:
Reads OK to me, as somebody who has never used decorators in Javascript (only Python and C#'s similar features). You should have somebody more familiar with decorators take a look too.

From the new user perspective, I expected to see two things, though:

1. An example of returning a replacement value (descriptor?), or else a discussion of why this is not useful -- the current examples are all side-effecting. This is the first thing I expected to see based on my Python experience. And the Descriptor terminology threw me off until I went and started reading the Javascript proposal.
2. An explanation that basically this is the same thing as the Javascript proposal, except "here is how to add types to it". I don't know whether we should repeat the explanation in the proposal.

pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
+# Introduction
+
+With the introduction of Classes in TypeScript and ES6, there now exist certain scenarios that require additional features to support annotating or modifying classes and class members.
```
DanielRosenwasser: Classes -> classes

ES6 -> ES2015

I don't know if this is necessarily an accurate motivating factor.

pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
...
+Decorators provide a way to add both annotations and a meta-programming syntax for class declarations and members.
+Decorators are a [stage 1 proposal](https://github.com/wycats/javascript-decorators/blob/master/README.md) for JavaScript and are available as an experimental feature of TypeScript.
+
+> NOTE&emsp; Decorators are an experimental feature that may change in future releases.
```
DanielRosenwasser: Just use a colon

pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
...
+
+To enable experimental support for decorators, you must enable the `experimentalDecorators` compiler option either on the command line or in your `tsconfig.json`:
+
+**Command Line**:
```
DanielRosenwasser: Can you make these into bulleted list items?

pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
...
+    "compilerOptions": {
+        "target": "ES5",
+        "experimentalDecorators": true
+    }
+}
+```
+
+# Decorators
+
+A *Decorator* is a special kind of declaration that can be attached to a [class declaration](#class-decorators), [method](#method-decorators), [accessor](#accessor-decorators), [property](#property-decorators), or [parameter](#parameter-decorators).
```
DanielRosenwasser: Decorator -> decorator

pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
...
+Decorators use the form `@expression`, where `expression` must evaluate to a function that will be called at runtime with information about the decorated declaration.
```
DanielRosenwasser: where `expression` is just a value that will be called at runtime.
When called, the function will be given information about the declaration being decorated.

pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
...
+
+For example, given the decorator `@sealed` we might write the `sealed` function as follows:
```
DanielRosenwasser: I don't think this is necessarily a useful line or example, since you aren't giving a specific implementation, so this is only reiterating that your decorator needs to be a function. A new user might still be lost with respect to the utility.

pages/Decorators.md
```diff
@@ -0,0 +1,472 @@
...
+
+```ts
+function sealed(target) {
+    // do something with "target" ...
+}
+```
+
+> NOTE&emsp; You can see a more detailed example of a decorator in [Class Decorators](#class-decorators), below.
```
DanielRosenwasser: Colon

pages/Decorators.md
DanielRosenwasser: Actually, we can look into this later.

DanielRosenwasser:
I think on the whole this is good; let's get the nits taken care of, pull it in, and and we can amend things afterwards.

mhegazy:
+1 for @DanielRosenwasser 's suggestion

PR #81, opened 2015-10-15
New Topic - Abstract Classes

### LINKED ISSUE #79: Abstract classes
Cover abstract classes

bricewilson:
Added a section explaining abstract classes and abstract methods along with sample code demonstrating their use.

---------

*Edit by @DanielRosenwasser*: Fixes #79.

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -77,6 +77,8 @@ The example also shows how to override methods in the base class with methods th
 Here both `Snake` and `Horse` create a `move` method that overrides the `move` from `Animal`, giving it functionality specific to each class.
 Note that even though `tom` is declared as an `Animal`, since its value is a `Horse`, when `tom.move(34)` calls the overriding method in `Horse`:
 
+Derived classes that contain constructor functions must call `super()` which will execute the constructor function on the base class.
+
 ```
 Slithering...
 Sammy the Python moved 5m.
@@ -292,46 +294,59 @@ console.log(grid2.calculateDistanceFromOrigin({x: 10, y: 10}));
 ```
 # Abstract Classes
 
-Abstract classes are base classes from which other classes may be derived. They may not be instantiated directly.
-They operate similar to abstract classes in other popular languages such as C# and Java. Unlike an interface, an
-abstract class may contain implementation details for it's members. The `abstract` keyword is used to define
-abstract classes as well as abstract methods within an abstract class.
-
-Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented
-in derived classes.
-
-Classes derived from abstract classes that also contain constructor functions must call `super()` which will execute
-the constructor function on the abstract base class.
+Abstract classes are base classes from which other classes may be derived.
+They may not be instantiated directly.
+Unlike an interface, an abstract class may contain implementation details for its members.
+The `abstract` keyword is used to define abstract classes as well as abstract methods within an abstract class.
 
 ```TypeScript
-abstract class DepartmentBase {
+abstract class Animal {
+    abstract makeSound(): void;
+    move(): void {
+        console.log('roaming the earth...');
+    }
+}
+```
+
+Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented in derived classes.
+Abstract methods share a similar syntax to interface methods.
+Both define the signature of a method without including a method body.
+However, abstract methods must include the `abstract` keyword and may optionally contain access modifiers.
+
+```TypeScript
+abstract class Department {
 	
-	constructor(public name: string) {
-	}
+    constructor(public name: string) {
+    }
 	
-	PrintName(): void {
-		console.log('Department name: ' + this.name);
-	}
+    printName(): void {
+        console.log('Department name: ' + this.name);
+    }
 	
-	abstract PrintMeeting(): void; // must be implemented in derived classes
+    abstract printMeeting(): void; // must be implemented in derived classes
 }
 
-class AccountingDepartment extends DepartmentBase {
+class AccountingDepartment extends Department {
 	
-	constructor() {
-		super('Accounting and Auditing'); // constructors in derived classes must call super()
-	}
+    constructor() {
+        super('Accounting and Auditing'); // constructors in derived classes must call super()
+    }
 	
-	PrintMeeting(): void {
-		console.log('The Accounting Department meets each Monday at 10am.');
-	}
+    printMeeting(): void {
+        console.log('The Accounting Department meets each Monday at 10am.');
+    }
+	
+    generateReports(): void {
+        console.log('Generating accounting reports...');
+    }
 }
 
-let dept: DepartmentBase = new DepartmentBase(); // error: cannot create an instance of an abstract class
-
-let acctDept: AccountingDepartment = new AccountingDepartment();
-acctDept.PrintName();
-acctDept.PrintMeeting();
+let department: Department; // ok to create a reference to an abstract type
+department = new Department(); // error: cannot create an instance of an abstract class
+department = new AccountingDepartment(); // ok to create and assign a non-abstract subclass
+department.printName();
+department.printMeeting();
+department.generateReports(); // error: method doesn't exist on declared abstract type
 ```
 
 # Advanced Techniques
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -311,7 +311,7 @@ abstract class Animal {
 Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented in derived classes.
 Abstract methods share a similar syntax to interface methods.
 Both define the signature of a method without including a method body.
-However, abstract methods must include the `abstract` keyword and may optionally contain access modifiers.
+However, abstract methods must include the `abstract` keyword and may optionally include access modifiers.
 
 ```TypeScript
 abstract class Department {
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
 console.log(grid1.calculateDistanceFromOrigin({x: 10, y: 10}));
 console.log(grid2.calculateDistanceFromOrigin({x: 10, y: 10}));
 ```
+# Abstract Classes
+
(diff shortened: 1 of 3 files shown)


pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
 console.log(grid1.calculateDistanceFromOrigin({x: 10, y: 10}));
 console.log(grid2.calculateDistanceFromOrigin({x: 10, y: 10}));
 ```
+# Abstract Classes
+
+Abstract classes are base classes from which other classes may be derived. They may not be instantiated directly.
+They operate similar to abstract classes in other popular languages such as C# and Java. Unlike an interface, an
+abstract class may contain implementation details for it's members. The `abstract` keyword is used to define
```
sandersn: it's -> its

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
 console.log(grid1.calculateDistanceFromOrigin({x: 10, y: 10}));
 console.log(grid2.calculateDistanceFromOrigin({x: 10, y: 10}));
 ```
+# Abstract Classes
+
+Abstract classes are base classes from which other classes may be derived. They may not be instantiated directly.
+They operate similar to abstract classes in other popular languages such as C# and Java. Unlike an interface, an
```
sandersn: I'd drop the "similar to C# and Java" sentence. We should only compare Typescript to other languages when it adds a lot of value (eg we copy its unique features, or our implementation works *exactly* like theirs).

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
...
+Classes derived from abstract classes that also contain constructor functions must call `super()` which will execute
+the constructor function on the abstract base class.
+
+```TypeScript
+abstract class DepartmentBase {
+	
+	constructor(public name: string) {
+	}
+	
+	PrintName(): void {
```
sandersn: camelCase: printName

sandersn: and elsewhere in this sample

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
...
+	}
+	
+	PrintMeeting(): void {
+		console.log('The Accounting Department meets each Monday at 10am.');
+	}
+}
+
+let dept: DepartmentBase = new DepartmentBase(); // error: cannot create an instance of an abstract class
+
+let acctDept: AccountingDepartment = new AccountingDepartment();
```
sandersn: maybe add an example of creating a variable of the base type?

```ts
let dept: DepartmentBase = new DepartmentBase(); // error: cannot create an instance of an abstract class
let accounting: DepartmentBase = new AccountingDepartment(); // OK
```

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
...
+abstract classes as well as abstract methods within an abstract class.
+
+Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented
+in derived classes.
+
+Classes derived from abstract classes that also contain constructor functions must call `super()` which will execute
+the constructor function on the abstract base class.
+
+```TypeScript
+abstract class DepartmentBase {
```
sandersn: I think `Department` is a better name, but it's up to you.

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
...
 ```
+# Abstract Classes
+
+Abstract classes are base classes from which other classes may be derived. They may not be instantiated directly.
+They operate similar to abstract classes in other popular languages such as C# and Java. Unlike an interface, an
+abstract class may contain implementation details for it's members. The `abstract` keyword is used to define
+abstract classes as well as abstract methods within an abstract class.
+
+Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented
+in derived classes.
```
sandersn: Can you add a sentence or two about the syntax, explaining that it

1. Doesn't include the method body.
2. Is the same as the interface syntax except with the `abstract` keyword (and it allows access modifiers, I think).

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
 console.log(grid1.calculateDistanceFromOrigin({x: 10, y: 10}));
 console.log(grid2.calculateDistanceFromOrigin({x: 10, y: 10}));
```
sandersn: I feel like this topic should go before static since it still talks about instances. What do you think?

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
...
+Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented
+in derived classes.
+
+Classes derived from abstract classes that also contain constructor functions must call `super()` which will execute
+the constructor function on the abstract base class.
+
+```TypeScript
+abstract class DepartmentBase {
+	
+	constructor(public name: string) {
```
sandersn: 4-spaces instead of tabs, please.

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
...
+
+Abstract classes are base classes from which other classes may be derived. They may not be instantiated directly.
+They operate similar to abstract classes in other popular languages such as C# and Java. Unlike an interface, an
+abstract class may contain implementation details for it's members. The `abstract` keyword is used to define
+abstract classes as well as abstract methods within an abstract class.
+
+Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented
+in derived classes.
+
+Classes derived from abstract classes that also contain constructor functions must call `super()` which will execute
```
sandersn: This is true of any class that derives from another. Can you move this explanation to the end of the first 'Inheritance' section? The example there is already careful to call `super` but doesn't explain that it's required.

pages/Classes.md
DanielRosenwasser: I think it's fine to put it here.

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
 console.log(grid1.calculateDistanceFromOrigin({x: 10, y: 10}));
 console.log(grid2.calculateDistanceFromOrigin({x: 10, y: 10}));
 ```
+# Abstract Classes
+
+Abstract classes are base classes from which other classes may be derived. They may not be instantiated directly.
```
DanielRosenwasser: @bricewilson can you split sentences onto their own lines?

pages/Classes.md
DanielRosenwasser: To add to @sandersn's feedback, I think it would be helpful to have a minimal abstract class above this sentence. 

Basically I'd have
1. One abstract method.
2. One concrete method.

An easy one could be borrowed from earlier in the chapter:

```TypeScript
abstract class Animal {
    abstract makeSound(): void;
    move(): void {
        // ...
    }
}
```

Give the reader a small taste of what they look like and then you can break it down.

pages/Classes.md
```diff
@@ -290,6 +290,49 @@ var grid2 = new Grid(5.0);  // 5x scale
...
+	constructor() {
+		super('Accounting and Auditing'); // constructors in derived classes must call super()
+	}
+	
+	PrintMeeting(): void {
+		console.log('The Accounting Department meets each Monday at 10am.');
+	}
+}
+
+let dept: DepartmentBase = new DepartmentBase(); // error: cannot create an instance of an abstract class
```
DanielRosenwasser: You don't need a type annotation here or below.

pages/Classes.md
DanielRosenwasser: :+1: for `Department`

pages/Classes.md
DanielRosenwasser: Discussed with @sandersn offline. Consider having this example:

```TypeScript
let department: Department; // okay to have a reference to an abstract type

department = new Department();           // error: can't create an instance of an abstract class
department = new AccountingDepartment(); // okay to create a non-abstract subclass

department.printName();
department.printMeeting();
// maybe add an accounting-specific method that does't work here.

```

pages/Classes.md
bricewilson: I don't have a strong opinion, so I guess I'll side with @DanielRosenwasser and leave it here.

bricewilson:
I'm not sure if I'm following proper pull request protocol - this is a bit new to me, but I'm trying to learn. I pushed one new commit with all of the suggested changes. Would one commit per change have been preferred?

pages/Classes.md
sandersn: Yeah, @DanielRosenwasser pointed out offline that the static topic is not likely to ever get very long, since we talk about merging elsewhere.

sandersn:
I don't think we have an established style, so anything is fine. 
Personally, I prefer one commit per logical chunk of work (if possible) like you have here.

pages/Classes.md
```diff
@@ -290,6 +292,62 @@ var grid2 = new Grid(5.0);  // 5x scale
...
+    move(): void {
+        console.log('roaming the earth...');
+    }
+}
+```
+
+Methods within an abstract class that are marked as abstract do not contain an implementation and must be implemented in derived classes.
+Abstract methods share a similar syntax to interface methods.
+Both define the signature of a method without including a method body.
+However, abstract methods must include the `abstract` keyword and may optionally contain access modifiers.
```
sandersn: minor nit: may optionally **include** access modifiers

sandersn:
Besides one minor wording change, looks ready to merge.

pages/Classes.md
bricewilson: No problem. Just pushed the change.

DanielRosenwasser:
Hey @bricewilson can you pull in the changes from master? It looks like you'll need to manually merge recent changes which addressed linting issues.

sandersn:
Nice, thanks for the lint fix in Namespaces and Modules, too!

PR #85, opened 2015-10-26
Add guidance for modules and namespaces

mhegazy:
Add content in http://stackoverflow.com/questions/30357634/how-do-i-use-namespaces-with-typescript-external-modules to modules.md

### DIFF
--- a/pages/Modules.md
+++ b/pages/Modules.md
@@ -12,10 +12,10 @@ Conversely, to consume a variable, function, class, interface, etc. exported fro
 
 Modules are declarative; the relationships between modules are specified in terms of imports and exports at the file level.
 
-Modules import one another using a module loader. 
-At runtime the module loader is responsible for locating and executing all dependencies of a module before executing it. 
+Modules import one another using a module loader.
+At runtime the module loader is responsible for locating and executing all dependencies of a module before executing it.
 Common module loaders used in JavaScript are [CommonJS module loader](https://en.wikipedia.org/wiki/CommonJS) for Node.js applications, and [require.js](http://requirejs.org/) for Web applications.
- 
+
 In TypeScript just as in ECMAScript 2015, any file containing a top-level `import` or `export` is considered a module.
 
 # Export
@@ -60,7 +60,7 @@ export { ZipCodeValidator as mainValidator };
 
 ## Re-exports
 
-Often modules extend other modules, and partially expose some of their features. 
+Often modules extend other modules, and partially expose some of their features.
 A re-export does not import it locally, or introduce a local variable.
 
 ##### ParseIntBasedZipCodeValidator.ts
@@ -68,7 +68,7 @@ A re-export does not import it locally, or introduce a local variable.
 ```ts
 export class ParseIntBasedZipCodeValidator {
     isAcceptable(s: string) {
-        return s.length === 5 && parseInt(s).toString() === s;;
+        return s.length === 5 && parseInt(s).toString() === s;
     }
 }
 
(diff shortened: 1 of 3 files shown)


pages/Modules.md
```diff
@@ -498,3 +498,269 @@ Now we can `/// <reference>` `node.d.ts` and then load the modules using `import
...
 var myUrl = URL.parse("http://www.typescriptlang.org");
 ```
+
+# Modules and Namespaces
+
+> For more discussion about modules and namespaces see [[Modules and Namespaces]]
+
+## How do I use namespaces with TypeScript modules
+
+> **Do not use "namespaces" in external modules. Don't do this. Seriously. Stop.**
```
sandersn: Are we still using the term external module? I thought they were now just called modules.

pages/Modules.md
```diff
@@ -498,3 +498,269 @@ Now we can `/// <reference>` `node.d.ts` and then load the modules using `import
...
+}
+```
+
+#### MyFunc.ts
+
+```ts
+export default function getThing() { return 'thing'; }
+```
+
+#### Consumption.ts
```
sandersn: Maybe Consumer instead of Consumption? Consumption sounds like a disease to me.

pages/Modules.md
```diff
@@ -498,3 +498,269 @@ Now we can `/// <reference>` `node.d.ts` and then load the modules using `import
...
+
+```ts
+import { SomeType, SomeFunc } from "./MyThings";
+var x = new SomeType();
+var y = someFunc();
+```
+
+### If you're exporting a large number of things, only then should you use the `namespace` keyword
```
sandersn: I'm confused. The example below doesn't use the namespace keyword. Is the difference that it uses `export module`?

pages/Modules.md
```diff
@@ -498,3 +498,269 @@ Now we can `/// <reference>` `node.d.ts` and then load the modules using `import
...
+        if (this.operator) {
+            this.memory = this.evaluateOperator(this.operator, this.memory, this.current);
+        }
+        else {
+            this.memory = this.current;
+        }
+        this.current = 0;
+    }
+
+    public handelChar(char: string) {
```
sandersn: Typo:handle

pages/Modules.md
```diff
@@ -498,3 +498,269 @@ Now we can `/// <reference>` `node.d.ts` and then load the modules using `import
...
+export function test(c: Calculator, input: string) {
+    for (let i = 0; i < input.length; i++) {
+        c.handelChar(input[i]);
+    }
+
+    console.log(`result of '${input}' is '${c.getResult()}'`);
+}
+```
+
+Here is a simple test for the calcultor using the exposed `test` function.
```
sandersn: Typo:calculator

pages/Modules.md
```diff
@@ -498,3 +498,269 @@ Now we can `/// <reference>` `node.d.ts` and then load the modules using `import
...
+
+```ts
+import { Calculator, test } from "./Calculator";
+
+
+var c = new Calculator();
+test(c, "1+2*33/11="); // prints 9
+```
+
+Now to extend this to a add support for input with numbers in bases other than 10, lets create `ProgrammerCalculator.ts`
```
sandersn: Remove extra 'a' and change to 'let's'

sandersn:
Some minor terminology and typos. Otherwise :+1:

DanielRosenwasser:
I don't think the writing style here meshes well with the handbook. It can't just be thrown in verbatim.

pages/Modules.md
mhegazy: done.

pages/Modules.md
mhegazy: done.

pages/Modules.md
mhegazy: done.

pages/Modules.md
mhegazy: done.

pages/Modules.md
mhegazy: i will replace the whole thing.

mhegazy:
@DanielRosenwasser i have changed the section about modules and namespaces so that it follows the same tone as the rest of the document. the changes are in https://github.com/Microsoft/TypeScript-Handbook/pull/89

PR #87, opened 2015-10-29
Basic description of tuples

### LINKED ISSUE #86: Add an out of date notice
The handbook currently doesn't mention things like tuples, and this makes it confusing if you learn about tuples from a different source and then try to look them up in the handbook. The handbook is also the main form of documentation in the TS website, so it's natural to expect it being up to date, which is an another reason to add a notice that it doesn't cover all the TS features.

sandersn:
Fixes #86.

Clone from "What's new in Typescript" into Basic Types. A more complete example should be put elsewhere, along with an explanation of destructuring. The handbook focusses almost exclusively on types and modules right now, since that's what Typescript adds. So there's not really a good place to talk about expressions right now.

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -72,7 +72,7 @@ var list: Array<number> = [1, 2, 3];
 # Tuple
 
 Tuple types allow you to express an array where the type of a fixed number of elements is known, but need not be the same.
-For example, you may want to represent an array with a `string` at position 0 and a `number` at position 1:
+For example, you may want to represent a value as a pair of a `string` and a `number`:
 
 ```TypeScript
 // Declare a tuple type
@@ -97,6 +97,8 @@ x[3] = 'world'; // OK, string can be assigned to (string | number)
 console.log(x[5].toString()); // OK, 'string' and 'number' both have toString
 x[6] = true; // Error, boolean isn't (string | number)
 ```
+
+Union types are an advanced topic that we'll cover in a later chapter.
 
 # Enum
 
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -69,6 +69,35 @@ The second way uses a generic array type, `Array<elemType>`:
 var list: Array<number> = [1, 2, 3];
 ```
 
+# Tuple
+
+Tuple types allow you to express an array where the type of a fixed number of elements is known, but need not be the same. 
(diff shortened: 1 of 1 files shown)


mhegazy:
:+1:

pages/Basic Types.md
```diff
@@ -69,6 +69,35 @@ The second way uses a generic array type, `Array<elemType>`:
 var list: Array<number> = [1, 2, 3];
 ```
 
+# Tuple
+
+Tuple types allow you to express an array where the type of a fixed number of elements is known, but need not be the same.
+For example, you may want to represent an array with a `string` at position 0 and a `number` at position 1:
```
DanielRosenwasser: represent a value as a pair of a `string` and a `number`

pages/Basic Types.md
```diff
@@ -69,6 +69,35 @@ The second way uses a generic array type, `Array<elemType>`:
...
+```
+
+When accessing an element with a known index, the correct type is retrieved:
+
+```TypeScript
+console.log(x[0].substr(1)); // OK
+console.log(x[1].substr(1)); // Error, 'number' does not have 'substr'
+```
+
+When accessing an element outside the set of known indices, a union type is used instead:
```
DanielRosenwasser: We'll need to eventually link to union types

pages/Basic Types.md
sandersn: done

pages/Basic Types.md
sandersn: done, although I'm not even sure of the value of covering the error case -- I just copied it from the release notes. We kind of need a tutorial-style handbook *and* a set of reference pages that explicate the spec (@weswigham's tutorial on migrating to ES2015 modules is another good candidate for reference documentation). I'll leave it in until we make a reference section, though.

DanielRosenwasser: What error case?

sandersn: I was thinking of a runtime error, which we do not catch at compile time..

```ts
var x: [string, number] = ['hello', 10];
console.log(x[5].toString()); 
// would have been OK if we assigned to x[5] earlier, but we didn't in this case.
```

Telling people that they can add additional entries and properties to a tuple is not the most basic use of them -- it falls out from their array and object inheritance.

DanielRosenwasser:
:+1:

PR #88, opened 2015-11-03
Fix typo: Unnecessary word

SH4DY:


sandersn:
Thanks, @SH4DY! Good to know somebody is reading the handbook with care. :)

PR #90, opened 2015-11-04
Support for getters/setters in applyMixins function.

dsherret:
Hello, with the `applyMixins` function I have made the following additions:

1. Added missing semi-colon.
2. Do not copy the constructor over.
3. Added support for getters/setters in addition to methods.


mhegazy:
I do not think this is needed. the idea is to explain how you can define types flowing in/out of a mix-in function, and not to provide a full-fledged mix-in implementation. the current implementation, though does not handle accessors, is simpler to read.

dsherret:
Ok, that makes sense if that's the goal. Might be good to still add the missing semi-colon though.

Please close this issue if you think it should be closed.

mhegazy:
Adding the semicolon is fine off course.

PR #91, opened 2015-11-05
Remove extra word

jseiden:
I removed the unneeded 'to' from line 22. 
This changes 'the compiler only checks **to that**..." to "the compiler only checks **that**..."

The original sentence in question: 
> Notice that our object actually has more properties than this, but the compiler only checks **to that**
> *at  least* the ones required are present and match the types required."


sandersn:
Thanks @jseiden!

PR #93, opened 2015-11-08
Fix typo: use minute instead of number

zhongsp:
One typo in Interfaces.md. It should be 
```ts
function createClock(ctor: ClockConstructor, hour: number, minute: number): ClockInterface {
    return new ctor(hour, minute);
}
```
Not
```ts
function createClock(ctor: ClockConstructor, hour: number, minute: number): ClockInterface {
    return new ctor(hour, number);
}
```


sandersn:
Thanks @zhongsp !

PR #96, opened 2015-11-09
Another small typo!

ColinEberhardt:


pages/Classes.md
```diff
@@ -230,7 +230,7 @@ if (employee.fullName) {
 }
 ```
 
-While allowing people to randomly set fullName directly is pretty handy, this might get us in trouble if we people can change names on a whim.
+While allowing people to randomly set fullName directly is pretty handy, this might get us in trouble if people can change names on a whim.
```
mhegazy: how about?
```md
while having the ability  to randomly set fullName directly is pretty handy, this might get us in trouble if we can change names on a whim.
```

ColinEberhardt:
Yes, I like that - referring to 'people' is rather ambiguous.

mhegazy:
Thanks!

PR #99, opened 2015-11-18
Updated content in 'Functions.md'

DanielRosenwasser:


pages/Functions.md
```diff
@@ -127,44 +127,64 @@ function buildName(firstName: string, lastName?: string) {
...
 ```ts
 function buildName(firstName: string, lastName?: string) {
+    // ...
+}
 ```
 
 and
 
 ```ts
 function buildName(firstName: string, lastName = "Smith") {
```
sandersn: I want to confirm because I haven't used this feature before: lastName doesn't need a trailing `?` here, like `lastName? = "Smith"`, right?

pages/Functions.md
```diff
@@ -127,44 +127,64 @@ function buildName(firstName: string, lastName?: string) {
...
+var result1 = buildName("Bob");                  // works correctly now, returns "Bob Smith"
+var result2 = buildName("Bob", undefined);       // still works, also returns "Bob Smith"
+var result3 = buildName("Bob", "Adams", "Sr.");  // error, too many parameters
+var result4 = buildName("Bob", "Adams");         // ah, just right
 ```
 
-Just as with optional parameters, default parameters must come after required parameters in the parameter list.
-
-Optional parameters and default parameters also share what the type looks like. Both:
+Default-initialized parameters that come after all required parameters are treated as optional, and just like optional parameters, can be omitted when calling their respective function.
```
sandersn: Never mind, I read it in the new paragraph. I think most people will have the same question when reading, so maybe the new paragraph can be integrated better into the existing text than just sticking it on the end.

pages/Functions.md
DanielRosenwasser: Confirmed.

PR #102, opened 2015-11-19
add missing declare

uwinkelvos:
would not compile otherwise:
"A 'declare' modifier is required for a top level declaration in a .d.ts file"


mhegazy:
thanks!

PR #104, opened 2015-11-21
fix a couple broken links

csantero:


DanielRosenwasser:
Thanks for the PR @csantero, but this was an intentional change until we figure things out with our process of converting from markdown to html (which might end up being a regular expression find/replace).

csantero:
@DanielRosenwasser gotcha, sorry about that

PR #110, opened 2015-11-26
x should be an Object

MagiCarbon:
x should be an Object to use *isAcceptable* method


DanielRosenwasser:
Thanks for the PR @MagiCarbon, but you're actually `new`-ing up `x.isAcceptable`, which would be `undefined`. You'd need to add parens following the `new x` (i.e. `new x()`).

Can you instead just make the bound name `ZipCodeValidator` and just create a separate variable named `x` and assign it a `new ZipCodeValidator()`?

MagiCarbon:
hello, i've added the parentheses. sorry about that mistake.

Issue #116, opened 2015-12-02
Use `let` as much as possible


RyanCavanaugh:
We wrote the majority of the handbook before the introduction of `let`, and `var` is stupid.

DanielRosenwasser:
`var` is stupid, but first we need a section introducing `let` and `const`.

DanielRosenwasser:
That issue I raised is #32. We can close this.

PR #119, opened 2015-12-02
Let

DanielRosenwasser:
Addresses #116.

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -21,7 +21,7 @@ In addition to hexadecimal and decimal literals, TypeScript also supports binary
 let decimal: number = 6;
 let hex: number = 0xf00d;
 let binary: number = 0b1010;
-let octal: number = 0o755;
+let octal: number = 0o744;
 ```
 
 # String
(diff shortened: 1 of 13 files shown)


pages/Basic Types.md
```diff
@@ -184,5 +187,11 @@ function warnUser(): void {
...
-```
\ No newline at end of file
+let unusable: void = undefined;
+```
+
+# A note about `let`
+
+You may've noticed that so far, we've been using the `let` keyword instead of JavaScript's `var` keyword which you might be more familiar with.
+The `let` keyword is actually a newer JavaScript construct that TypeScript makes available.
+We'll discuss the details later, but many common problems in JavaScript are alleviated by using `let`, so it's best practice to adopt it over using `var` whenever possible.
```
mhegazy: except that later never happens :)

mhegazy: `let` and `const` ?

mhegazy:
:+1:

pages/Basic Types.md
DanielRosenwasser: That's future Daniel's problem, and he's a jerk anyhow.

PR #122, opened 2015-12-03
Added a section on 'var'/'let'/'const'.

### LINKED ISSUE #32: let and const
Scoping rules, differences between var, let and const.

DanielRosenwasser:
Fixes #32.

### DIFF
--- a/pages/Variable Declarations.md
+++ b/pages/Variable Declarations.md
@@ -52,19 +52,20 @@ Even if `g` is called once `f` is done running, it will be able to access and mo
 
 ```ts
 function f() {
-    var a = 10;
-    a = 15;
+    var a = 1;
+
+    a = 2;
     var b = g();
-    a = 20;
+    a = 3;
+
     return b;
 
     function g() {
-        var c = a + 1;
-        return c;
+        return a;
     }
 }
 
-f(); // returns 16
+f(); // returns 2
 ```
 
 ## Scoping rules
@@ -113,7 +114,7 @@ As experienced developers know by now, similar sorts of bugs slip through code r
 
 ## Variable capturing quirks
 
-Take a quick second to guess whta the output of the following snippet is:
+Take a quick second to guess what the output of the following snippet is:
 
 ```ts
 for (var i = 0; i < 10; i++) {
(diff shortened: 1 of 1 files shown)


pages/Variable Declarations.md
```diff
@@ -0,0 +1,422 @@
...
+```ts
+function f() {
+    var a = 10;
+    a = 15;
+    var b = g();
+    a = 20;
+    return b;
+
+    function g() {
+        var c = a + 1;
```
mhegazy: c is not interesting here, just return `a`

pages/Variable Declarations.md
```diff
@@ -0,0 +1,422 @@
...
+    return sum;
+}
+```
+
+Maybe it was easy to spot out for some, but the inner `for`-loop will accidentally overwrite the variable `i` because `i` refers to the same function-scoped variable.
+As experienced developers know by now, similar sorts of bugs slip through code reviews and can be an endless source of frustration.
+
+## Variable capturing quirks
+
+Take a quick second to guess whta the output of the following snippet is:
```
mhegazy: typo, what

pages/Variable Declarations.md
```diff
@@ -0,0 +1,422 @@
...
+    let x = 100;
+    var x = 100; // error: can't have both declarations of 'x'
+}
+```
+
+That's not to say that block-scoped variable can never be declared with a function-scoped variable.
+The block-scoped variable just needs to be declared within a distinctly different block.
+
+```ts
+function f(condition, x) {
```
mhegazy: since we are talking about var and let, i would make x a local var declaration instead.

pages/Variable Declarations.md
```diff
@@ -0,0 +1,422 @@
...
+It is a bit of a double-edged sword in that it can introduce certain bugs on its own in the event of accidental shadowing, while also preventing certain bugs.
+For instance, imagine we had written our earlier `sumMatrix` function using `let` variables.
+
+```ts
+function sumMatrix(matrix: number[][]) {
+    let sum = 0;
+    for (let i = 0; i < matrix.length; i++) {
+        var currentRow = matrix[i];
+        for (let i = 0; i < currentRow.length; i++) {
+            sum += currentRow[i];
```
mhegazy: i would add a comment `// use i from inner loop

pages/Variable Declarations.md
```diff
@@ -0,0 +1,422 @@
...
+Like most broad questions, the answer is: it depends.
+
+Applying the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege), all declarations other than those you plan to modify should use `const`.
+The rationale is that if a variable didn't need to get written to, others working on the same codebase shouldn't automatically be able to write to the object, and will need to consider whether they really need to reassign to the variable.
+Using `const` also makes code more predictable when reasoning about flow of data.
+
+On the other hand, `let` is not any longer to write out than `var`, and many users will prefer its brevity.
+The majority of this handbook uses `let` declarations in that interest.
+
+Use your best judgement and discuss it with whoever else will be contributing to your codebase.
```
mhegazy: `and consult with your team coding standards`?

mhegazy:
this is great, thanks. one thing to add is a action about Temporal Dead Zone (TDZ). there is a good write up for it in http://jsrocks.org/2015/01/temporal-dead-zone-tdz-demystified/

PR #124, opened 2015-12-05
Fix wording of interface naming conventions, fix #121

### LINKED ISSUE #121: Prohibition against prefixing interfaces with "I"
The Typescript Handbook tells the developer not to prefix interfaces with I. This needs more detail. Lots of .NET developers are familiar with I-prefixed interfaces. Simply stating that the concept is broader is not sufficient. If it's an antipattern, let's describe why. If it's not an antipattern, then let's say "it is not necessary to prefix interfaces with I, and Typescript Contributors do X"

zhongsp:
Update the wording of interface naming conventions in TypeScript, so users won't be confused.

### DIFF
--- a/pages/Writing Definition Files.md
+++ b/pages/Writing Definition Files.md
@@ -93,7 +93,7 @@ The trade-offs here are as follows:
 
 ## Naming Conventions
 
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Thanks @zhongsp!

PR #132, opened 2015-12-14
Adding consistency with examples used elsewhere in docs and to render nicer in HTML

bgashler1:
I changed all example code sections I could find to have their title set at headline level 5, so that they don't draw a lot of attention to people scanning the page quickly.  Additionally for filenames, I changed them to not have code elements, since filenames are not code per se.

I also removed unordered lists that were causing the markdown renderer on the website to make code fences not render as code blocks in HTML.

### DIFF
--- a/pages/Modules.md
+++ b/pages/Modules.md
@@ -523,7 +523,7 @@ If a module's primary purpose is to house one specific export, then you should c
 This makes both importing and actually using the import a little easier.
 For example:
 
-#### MyClass.ts
+##### MyClass.ts
 
 ```ts
 export default class SomeType {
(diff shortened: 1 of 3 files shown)


DanielRosenwasser:
The unordered list thing is a separate issue which I have another fix for. I also don't necessarily think that using H5s is necessarily the correct thing to do here. I'm going to push out my change and we can consider this afterwards.

mhegazy:
@DanielRosenwasser can you link your changes? is this still an issue?

DanielRosenwasser:
I did. I don't think we should be changing code samples to use H5s, so I think this can be closed.

PR #135, opened 2015-12-21
Fix typo

sdenel:


sandersn:
@sdenel, I think this is correct as-is. There is an elided word, but it is 'use'.

> Once defined, we can use this function type interface like we would *use* other interfaces.

PR #136, opened 2016-01-05
Fix an error in advanced types example

zhongsp:
Fix an error in code sample.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -277,7 +277,7 @@ type XYZCoord = { x: XCoord; y: YCoord; z: number };
 type Coordinate = XCoord | XYCoord | XYZCoord;
 type CoordList = Coordinate[];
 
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Nice catch! Though, I wonder if this is just confusing. @sandersn @mhegazy, do you two have any thoughts on this?

sandersn:
Yes, it's confusing. Union types are already discussed above, so I'd just delete everything in the code block from `type XYZCoord` and below.

sandersn:
The union example is not useful in addition -- it's a jumble of types that's really hard to distinguish with a type guard, at least without some kind of type tag which doesn't work well at the moment.

zhongsp:
@DanielRosenwasser The update in PR 138 is more clear. I think we could close this PR. :smile:

mhegazy:
thanks @zhongsp!

PR #138, opened 2016-01-05
Remove confusing alias example

sandersn:
Union types are already explained, and the types that are aliased are next to impossible to distinguish with a type guard to boot.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -271,6 +271,9 @@ Type aliases are sometimes similar to interfaces, but can name primitives, union
 type XCoord = number;
 type YCoord = number;
 type XYCoord = { x: XCoord; y: YCoord };
+let x: XCoord = 10;
+let y: YCoord = x;
+let coords: XYCoord = { x, y };
 ```
 
 Aliasing doesn't actually create a new type - it creates a new *name* to refer to that type.
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -280,16 +280,20 @@ Type aliases create a new name for a type.
 Type aliases are sometimes similar to interfaces, but can name primitives, unions, tuples, and any other types that you'd otherwise have to write by hand.
 
 ```ts
-type XCoord = number;
-type YCoord = number;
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
I'd prefer we actually showed an example where we show something that has the type of an alias, so I don't think removing it entirely is the right move. Using different names would be more optimal.

sandersn:
A value of the aliased type, right? I'll put one back in, just not a value of a union type.

DanielRosenwasser:
That's better, but we now need something to show a type alias for a union.

sandersn:
Why? They are orthogonal concepts and showing that they work together just makes the explanation over-complicated.

mhegazy:
what is the status of this PR? 
I think there is value in showing that type aliases can alias unions, intersection, and arrays. may be add a function that accepts an array of Coordinates defined as a type alias.

sandersn:
Here's the example I came up with

```ts
type Coord = XCoord | YCoord | XYCoord;
function distance(c: Coord): number {
    if (typeof c === 'number') {
        return c;
    }
    else {
        return Math.sqrt(Math.pow(c.x, 2) + Math.pow(c.y, 2));
    }
}
```

It's still a bit wonky because XCoord and YCoord are really the same type, and writing other functions for this (or similar data structures) would break down pretty fast. For example:

```
// how to distinguish X from Y?
declare function xComponent(c: Coord): number;
// how to distinguish XY from Polar?
type PolarCoord = { angle: number, radius: number }
type Coord = X | Y | XY | Polar;
```

So it's not quite a real example still.

mhegazy:
how about `string | (()=> string)`, something like 

```ts
type Name = string;
type NameResolver = ()=> Name;
type NameOrNameResolver = Name | NameResolver;

function getName(n: NameOrNameResolver): Name {
    if (typeof n === "string") {
      return n;
   }
   else {
      return n();
   }
}
```

sandersn:
Good idea. I like the Name example better.

mhegazy:
:+1:

PR #142, opened 2016-01-17
Add example typing for "Function Module" pattern

### LINKED ISSUE #141: Document definition file pattern for ambient modules which export single function.
This is a common pattern for node modules, and it wasn't immediately apparent to me from the "Single Complex Object in External Modules" example.

Example: https://www.npmjs.com/package/stream-to-promise

```typescript
declare module 'stream-to-promise' {
  import {Promise} from 'es6-promise';
  export default function streamToPromise<T>(stream: NodeJS.ReadableStream): Promise<T>;
}
```

traviskaufman:
Many node modules such as rimraf, debug, mkdirp, etc. export a single
function by default that can then be used by clients. This commit adds
an example of writing a typing for this pattern. Node module authors
will be able to look at this example and immediately understand how to
write a type defintion for their package.

Fixes #141.

### DIFF
--- a/pages/Writing Definition Files.md
+++ b/pages/Writing Definition Files.md
@@ -230,6 +230,25 @@ declare var Eagle: Eagle;
 export = Eagle;
 ```
 
+## Module as a Function
+
+#### Usage
+
+```ts
+// Common pattern for node modules (e.g. rimraf, debug, request, etc.)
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Thanks!

PR #145, opened 2016-02-07
Classes: Use more obvious syntax for typeof.

omeid:
It is not obvious at first glance what is exactly going on. I think putting parenthesis there makes it more clear.

Not sure if it needs to be mention that the parenthesis are optional.

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -422,7 +422,7 @@ let greeter1: Greeter;
 greeter1 = new Greeter();
 console.log(greeter1.greet());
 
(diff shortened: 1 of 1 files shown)


sandersn:
I don't think the parentheses are useful here -- parentheses are usually there to help visual parsing, but we want new users to quickly learn that `: ... =` are the usual boundaries for types, not `( ... )`. Plus, I think the surrounding text scaffolds this new syntax adequately so that you should be primed to see something new in the type annotation.

PR #146, opened 2016-02-08
Typo fix: Defualt -> Default

NKCSS:
Typo fix: Defualt -> Default


DanielRosenwasser:
Thanks @NKCSS!

NKCSS:
You're welcome @DanielRosenwasser :)

N.B.

I noticed there's quite some difference between the version that's viewable online (http://www.typescriptlang.org/Handbook) and the one here in the git repo; how/when is that updated?

PR #152, opened 2016-02-16
Added a small section on type assertions.

### LINKED ISSUE #123: Type assertions are not explained (despite being used)
Tonight I looked at the Typescript handbook (it was linked on typescriptlang.org). I was specifically wanting to understand the meaning of a piece of code I had found on StackOverflow: `var canvas = <HTMLCanvasElement> document.getElementById("mycanvas");`

When I looked in the handbook-- the only piece of user-facing documentation for Typescript right now, it looks like-- I found the angle brackets were not documented. Weirder, very early in the handbook, the angle brackets are *used*. However they are not at that point explained. I want to highlight the code:

    interface Shape {
        color: string;
    }

    interface Square extends Shape {
        sideLength: number;
    }

    let square = <Square>{};
    square.color = "blue";
    square.sideLength = 10;

The \<\> is obviously some kind of cast operator. However, after looking at this code I have many questions:

- Does \<\> merely type-check? Or does it coerce (convert)?
- Is \<\> safe? What happens if the conversion is illegal?
- Can I use \<\> on "primitive" types such as number and string?

A friend helped me find section 4.16 of the spec (which is hard to find because it is named "type assertions", not something familiar like "casting"), which sort of explains \<\>, but I am still a little confused. This really should be explained in the handbook.

I think when you drop the above code in, since it is the first use of the \<\> construct, you should take the time to explain it. I think it is especially important to explain it because this code seems to do something "funny". That is, it looks to me like that code is doing something illegal. `let square:Square = {};` is not legal, because color and sideLength are non-optional. But the \<Square\> allows it to occur, even though it is unsafe and code using could throw an exception. If explained correctly, this helps the user understand an important element of writing Typescript (that \<\>s violate the typesystem and make your program unsafe). However if not explained, it could (1) teach readers of the handbook bad practice-- since you're flat out instructing them in a pattern which subverts the typesystem; and (2) create confusion when reading other parts of the guide (I had trouble understanding the "optional types" section when I read it because I had first read this bit of early sample code where two non-optional types are not specified).

DanielRosenwasser:
Fixes #123.

@sandersn @BrendanShah @bowdenk7

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -192,20 +192,19 @@ let unusable: void = undefined;
 
 # Type assertions
 
-Sometimes you'll end up in a situation where you'll know a little better than TypeScript about the type of something.
-Usually, this will happen when you know the type of a value is more specific than its current type.
+Sometimes you'll end up in a situation where you'll know more about a value than TypeScript does.
+Usually this will happen when you know the type of some entity could be more specific than its current type.
 
 *Type assertions* are a way to tell the compiler "trust me, I know what I'm doing."
-A type assertion is like a type cast in other languages, but where some languages might perform some special checking or restructuring of data, a type assertion has no runtime impact.
-The assumption is that you, the programmer, have performed any special checks that you need.
+A type assertion is like a type cast in other languages, but performs no special checking or restructuring of data.
+It has has no runtime impact, and is used purely by the compiler.
+TypeScript assumes that you, the programmer, have performed any special checks that you need.
 
 Type assertions have two forms.
 One is the "angle-bracket" syntax:
 
 ```ts
-let someValue: any;
-
-// ...
+let someValue: any = "this is a string";
 
 let strLength: number = (<string>someValue).length;
 ```
@@ -213,9 +212,7 @@ let strLength: number = (<string>someValue).length;
 And the other is the `as`-syntax:
 
 ```ts
-let someValue: any;
-
-// ...
+let someValue: any = "this is a string";
 
 let strLength: number = (someValue as string).length;
 ```
@@ -227,4 +224,4 @@ Using one over the other is mostly a choice of preference; however, when using T
 
 You may've noticed that so far, we've been using the `let` keyword instead of JavaScript's `var` keyword which you might be more familiar with.
 The `let` keyword is actually a newer JavaScript construct that TypeScript makes available.
-We'll discuss the details later, but many common problems in JavaScript are alleviated by using `let`, so it's a best practice to adopt it over using `var` whenever possible.
+We'll discuss the details later, but many common problems in JavaScript are alleviated by using `let`, so you should use it instead of `var` whenever possible.
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -190,8 +190,41 @@ Declaring variables of type `void` is not useful because you can only assign `un
 let unusable: void = undefined;
 ```
 
+# Type assertions
+
+Sometimes you'll end up in a situation where you'll know a little better than TypeScript about the type of something.
+Usually, this will happen when you know the type of a value is more specific than its current type.
+
+*Type assertions* are a way to tell the compiler "trust me, I know what I'm doing."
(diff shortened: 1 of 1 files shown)


pages/Basic Types.md
```diff
@@ -190,8 +190,41 @@ Declaring variables of type `void` is not useful because you can only assign `un
 let unusable: void = undefined;
 ```
 
+# Type assertions
+
+Sometimes you'll end up in a situation where you'll know a little better than TypeScript about the type of something.
+Usually, this will happen when you know the type of a value is more specific than its current type.
+
+*Type assertions* are a way to tell the compiler "trust me, I know what I'm doing."
+A type assertion is like a type cast in other languages, but where some languages might perform some special checking or restructuring of data, a type assertion has no runtime impact.
```
sandersn: you mention "other languages" twice. I'd reword to "A type assertion is like a type cast in other languages, but performs no special checking or restructuring of data. It has no runtime impact at all -- it is information for the compiler only."

pages/Basic Types.md
```diff
@@ -190,8 +190,41 @@ Declaring variables of type `void` is not useful because you can only assign `un
 let unusable: void = undefined;
 ```
 
+# Type assertions
+
+Sometimes you'll end up in a situation where you'll know a little better than TypeScript about the type of something.
```
sandersn: This is pretty wordy. How about "Sometimes you'll have more information about an object than TypeScript does." Or maybe "an object's type than TypeScript does".

pages/Basic Types.md
```diff
@@ -190,8 +190,41 @@ Declaring variables of type `void` is not useful because you can only assign `un
...
+Usually, this will happen when you know the type of a value is more specific than its current type.
```
sandersn: remove comma after "Usually"

pages/Basic Types.md
```diff
@@ -190,8 +190,41 @@ Declaring variables of type `void` is not useful because you can only assign `un
...
+
+*Type assertions* are a way to tell the compiler "trust me, I know what I'm doing."
+A type assertion is like a type cast in other languages, but where some languages might perform some special checking or restructuring of data, a type assertion has no runtime impact.
+The assumption is that you, the programmer, have performed any special checks that you need.
```
sandersn: "TypeScript assumes that you, ..."

sandersn: Or maybe "The compiler assumes that you, ..."

pages/Basic Types.md
```diff
@@ -190,8 +190,41 @@ Declaring variables of type `void` is not useful because you can only assign `un
...
+
+Type assertions have two forms.
+One is the "angle-bracket" syntax:
+
+```ts
+let someValue: any;
```
sandersn: I think it would be easier to understand if it had an initial string value.

pages/Basic Types.md
```diff
@@ -190,8 +190,41 @@ Declaring variables of type `void` is not useful because you can only assign `un
...
+
+The two samples are equivalent.
+Using one over the other is mostly a choice of preference; however, when using TypeScript with JSX, only `as`-style assertions are allowed.
+
 # A note about `let`
 
 You may've noticed that so far, we've been using the `let` keyword instead of JavaScript's `var` keyword which you might be more familiar with.
 The `let` keyword is actually a newer JavaScript construct that TypeScript makes available.
-We'll discuss the details later, but many common problems in JavaScript are alleviated by using `let`, so it's best practice to adopt it over using `var` whenever possible.
+We'll discuss the details later, but many common problems in JavaScript are alleviated by using `let`, so it's a best practice to adopt it over using `var` whenever possible.
```
sandersn: If we're going for folksy/plain, then why not just say "... by using `let`, so you should use `let` instead of `var` whenever possible."

Best practise sounds like business jargon.

DanielRosenwasser:
Ping

sandersn:
:+1:

mhegazy:
I think this deserves a page by itself.

PR #158, opened 2016-02-18
Add protected constructor topic

AbubakerB:
This PR adds some documentation for private and protected constructor in Classes.md 
([TypeScript-6885#comment](https://github.com/Microsoft/TypeScript/pull/6885#issuecomment-184949330))

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -109,7 +109,7 @@ class Animal {
 
 ## Understanding `private`
 
-When a constructor or member is marked `private`, it cannot be accessed from outside of its containing class. For example:
+When a member is marked `private`, it cannot be accessed from outside of its containing class. For example:
 
 ```ts
 class Animal {
@@ -117,13 +117,7 @@ class Animal {
     constructor(theName: string) { this.name = theName; }
 }
 
-class Employee {
-    private name: string;
-    private constructor(theName: string) { this.name = theName; }
-}
-
 new Animal("Cat").name; // Error: 'name' is private;
-new Employee("Bob"); // Error: 'Employee' is private;
 ```
 
 TypeScript is a structural type system.
@@ -195,12 +189,13 @@ console.log(howard.name); // error
 
 Notice that while we can't use `name` from outside of `Person`, we can still use it from within an instance method of `Employee` because `Employee` derives from `Person`.
 
-A constructor may also be marked `protected`. This means that the class cannot be accessed outside of its containing class, but can be extended. For example,
+A constructor may also be marked `protected`. 
+This means that the class cannot be instantiated outside of its containing class, but can be extended. For example,
 
 ```ts
 class Person {
     protected name: string;
-    protected constructor(name: string) { this.name = name; }
+    protected constructor(theName: string) { this.name = theName; }
 }
 
 // Employee can extend Person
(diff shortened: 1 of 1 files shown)


pages/Classes.md
```diff
@@ -117,7 +117,13 @@ class Animal {
     constructor(theName: string) { this.name = theName; }
 }
 
+class Employee {
+    private name: string;
+    private constructor(theName: string) { this.name = theName; }
+}
+
 new Animal("Cat").name; // Error: 'name' is private;
+new Employee("Bob"); // Error: 'Employee' is private;
```
DanielRosenwasser: The constructor of 'Employee' is private.

pages/Classes.md
```diff
@@ -189,6 +195,32 @@ console.log(howard.name); // error
 
 Notice that while we can't use `name` from outside of `Person`, we can still use it from within an instance method of `Employee` because `Employee` derives from `Person`.
 
+A constructor may also be marked `protected`. This means that the class cannot be accessed outside of its containing class, but can be extended. For example,
```
DanielRosenwasser: Place sentences on different lines.

pages/Classes.md
```diff
@@ -189,6 +195,32 @@ console.log(howard.name); // error
...
+    }
+}
+
+let howard = new Employee("Howard", "Sales");
+let john = new Person("John"); // Error: 'Person' is protected and only accessible within it's class.
+```
+
 ## Parameter properties
 
 In our last example, we had to declare a private member `name` and a constructor parameter `theName`, and we then immediately set `name` to `theName`.
```
DanielRosenwasser: This is no longer the last example

pages/Classes.md
```diff
@@ -189,6 +195,32 @@ console.log(howard.name); // error
...
+        this.department = department;
+    }
+
+    public getElevatorPitch() {
+        return `Hello, my name is ${this.name} and I work in ${this.department}.`;
+    }
+}
+
+let howard = new Employee("Howard", "Sales");
+let john = new Person("John"); // Error: 'Person' is protected and only accessible within it's class.
```
DanielRosenwasser: it's -> its

DanielRosenwasser: The 'Person' constructor is protected

pages/Classes.md
DanielRosenwasser: Don't end these with semicolons

pages/Classes.md
```diff
@@ -109,7 +109,7 @@ class Animal {
 
 ## Understanding `private`
 
-When a member is marked `private`, it cannot be accessed from outside of its containing class. For example:
+When a constructor or member is marked `private`, it cannot be accessed from outside of its containing class. For example:
```
mhegazy: i would move this to a section by itself. a private constructor is not the same as a private property. a class with private constructor can not be extended. so i would add a new paragraph or it, explaining that it can not be called nor extended, and add a static factory function to explain how this pattern would be used.

DanielRosenwasser: I agree with Mohamed. I think that this warrants a small section on its own. If you want to open up a separate PR, that might be easier.

pages/Classes.md
mhegazy: not "accessed" but rather "instantiated"

pages/Classes.md
AbubakerB: Sure, i'll make this PR dedicated to `protected` then :)

pages/Classes.md
AbubakerB: It seems like that's been the case for quite some time.
Anywho, done!

pages/Classes.md
AbubakerB: Done

pages/Classes.md
DanielRosenwasser: Don't worry about taking care of private. I'll adapt the current content for it.

DanielRosenwasser:
I've created a new branch called `release-2.0` where the content is available.

PR #159, opened 2016-02-19
React-webpack guide: list react/react-dom as externals 

liminzhu:
webpack.config.js in [react-webpack guide](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/quick-start/react-webpack.md#create-a-webpack-configuration-file) should list react and react-dom as externals otherwise the library code gets packed into your bundle.js, which usually is not what you want.

### DIFF
--- a/pages/quick-start/react-webpack.md
+++ b/pages/quick-start/react-webpack.md
@@ -167,7 +167,7 @@ module.exports = {
             { test: /\.js$/, loader: "source-map-loader" }
         ]
     },
-    
+
     externals: {
         "react": "React",
         "react-dom": "ReactDOM",
--- a/pages/quick-start/react-webpack.md
+++ b/pages/quick-start/react-webpack.md
@@ -117,6 +117,8 @@ Create a file at the root of `proj` named `index.html` with the following conten
     </head>
     <body>
         <div id="example"></div>
+        <script src="./node_modules/react/dist/react.js"></script>
+        <script src="./node_modules/react-dom/dist/react-dom.js"></script>
         <script src="./dist/bundle.js"></script>
     </body>
 </html>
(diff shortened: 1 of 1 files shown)


liminzhu:
@DanielRosenwasser

pages/quick-start/react-webpack.md
```diff
@@ -164,6 +166,11 @@ module.exports = {
             // All output '.js' files will have any sourcemaps re-processed by 'source-map-loader'.
             { test: /\.js$/, loader: "source-map-loader" }
         ]
+    },
+
```
DanielRosenwasser: Remove trailing spaces here

pages/quick-start/react-webpack.md
```diff
@@ -117,6 +117,8 @@ Create a file at the root of `proj` named `index.html` with the following conten
     </head>
     <body>
         <div id="example"></div>
+        <script src="./node_modules/react/dist/react.js"></script>
+        <script src="./node_modules/react-dom/dist/react-dom.js"></script>
```
DanielRosenwasser: This isn't right; if you deployed a website, then this directory wouldn't be available.

Do react.js and react-dom.js end up in your `dist` directory? If not, I'll need to add an entry about adding the libraries separately.

DanielRosenwasser: Will webpack's loader even pick these up correctly?

liminzhu: I don't think webpack cares where the externals are as webpack does not them. 

I just tried moving react/react-dom into ./dist and changed path in html; everything worked out fine :smile: .

DanielRosenwasser: Ah I see now that it will use the values of the externals object as general globals. That's interesting. I'll look into this a little bit more tomorrow.

DanielRosenwasser:
@liminzhu I've opened up an alternate PR (#176). If you could leave feedback on that I'd appreciate it!

liminzhu:
That looks good @DanielRosenwasser . Sorry for the late response.

PR #161, opened 2016-02-19
Add pull request template.

DanielRosenwasser:
**Fixes** #162.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,17 @@
+<!--
+Thank you for submitting a pull request!
+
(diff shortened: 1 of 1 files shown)


sandersn:
Does `**Fixes**: ` match the auto-close regex used by github? I don't think it does, and I've been meaning to update the one in the typescript repo so that it does too.

DanielRosenwasser:
@sandersn fixed.

sandersn:
Yeah, it looks like the regex is pretty much `/[Ff]ixes\s+#\d+/`. I have to assume it's reading the raw markdown too.

PR #168, opened 2016-02-27
removed duplicate word

polshgiant:
fixed typo

(new to git, sorry if I've done something incorrectly)


DanielRosenwasser:
Thanks for the fix!

PR #169, opened 2016-02-27
Use modules and properties in the React quick start guide.

DanielRosenwasser:
@sandersn @RyanCavanaugh @mhegazy

### DIFF
--- a/pages/quick-start/react-webpack.md
+++ b/pages/quick-start/react-webpack.md
@@ -125,8 +125,8 @@ ReactDOM.render(
 ```
 
 We just imported our `Hello` component into `index.tsx`.
-Notice we used a path *relative to `index.tsx`*.
-If we hadn't, TypeScript would've searched our `node_modules` folder.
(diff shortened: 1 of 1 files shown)


pages/quick-start/react-webpack.md
```diff
@@ -84,28 +93,42 @@ This command will create a file called `typings.json` and a folder called `typin
...
+export class Hello extends React.Component<HelloProps, {}> {
     render() {
-        return <h1>Hello from TypeScript and React!</h1>;
+        return <h1>Hello from {this.props.compiler} and {this.props.framework}!</h1>;
     }
 }
+```
+
+Note that while this example is quite *classy*, we didn't need to use a class.
+Other methods of using React (like functional stateless components) should work just as well.
```
mhegazy: we should add a link here to some additional documentation.

mhegazy:
:+1:

pages/quick-start/react-webpack.md
DanielRosenwasser: Good idea.

PR #170, opened 2016-02-27
Replace s/argument/parameter/i here and there

masak:
Commonly in programming languages (and specifically in the
TypeScript specification), "parameter" is used to describe a
variable accepted by a function, and "argument" describes a
value passed to a function during a call. Similarly, in the
case of generic types, "(type) parameter" is a part of the
type declaration and "(type) argument" a part of the use.

This commit fixes up a few cases where it seems "parameter" was
intended rather than "argument".


mhegazy:
thanks!

PR #176, opened 2016-03-03
Use 'externals' field in webpack file, use external dependencies, suggest CDN.

DanielRosenwasser:
Subsumes #159.

@sandersn @RyanCavanaugh @mhegazy

### DIFF
--- a/pages/quick-start/react-webpack.md
+++ b/pages/quick-start/react-webpack.md
@@ -152,8 +152,8 @@ Create a file at the root of `proj` named `index.html` with the following conten
 ```
 
 Notice that we're including files from within `node_modules`.
-React and React-DOM's npm packages include standalone `.js` files that you can include in a web page.
-Feel free to copy these files to another direction, or alternatively, host them on a content delivery network (CDN).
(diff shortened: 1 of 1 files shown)


pages/quick-start/react-webpack.md
```diff
@@ -140,11 +140,22 @@ Create a file at the root of `proj` named `index.html` with the following conten
     </head>
     <body>
         <div id="example"></div>
+
+        <!-- Dependencies -->
+        <script src="./node_modules/react/dist/react.js" />
```
mhegazy: should that be a CDN somewhere?

DanielRosenwasser: It doesn't matter where you get it from. I explain that below.

mhegazy: oh.. i see the comment below.

mhegazy: can not we add the CDN reference, and then tell them they can copy from node_modules instead?

what is the standard react tutorial does?

mhegazy:
:+1:

PR #177, opened 2016-03-05
Add a quick-start guide on Knockout.js

DanielRosenwasser:


pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+```ts
+declare var require: any;
+require.config({
+    paths: {
+        "knockout": "externals/knockout-3.4.0",
+    }
+});
+```
+
+This file is going to inform RequireJS of where to find Knockout when we import it like we did in `hello.ts`.
```
sandersn: cut out some words: "This file will tell RequireJS where to find Knockout ..."

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+ko.applyBindings(new HelloViewModel("TypeScript", "Knockout"));
+```
+
+Next, we'll create a file named `config.ts` in `src` as well.
+
+```ts
+declare var require: any;
```
mhegazy: this looks like a good place to talk about how to consume external dependencies in TS. and why you chose `any` vs a better definition, and may be a link to writing definition files in the handbook.

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+    constructor(language: string, framework: string) {
+        this.language = ko.observable(language);
+        this.framework = ko.observable(framework);
+    }
+}
+
+ko.applyBindings(new HelloViewModel("TypeScript", "Knockout"));
+```
+
+Next, we'll create a file named `config.ts` in `src` as well.
```
mhegazy: i think we need a note about what is require/a loader does in general. we usually get these questions. Users new to modules, do not find these concepts intuitive, so anything we do here will save them time later on.

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+```ts
+declare var require: any;
+require.config({
+    paths: {
+        "knockout": "externals/knockout-3.4.0",
+    }
+});
+```
+
+This file will tell RequireJS where to find Knockout when we import it, just like we did in `hello.ts`.
```
mhegazy: i would add a link here to http://requirejs.org/docs/api.html#config

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+            Hello from
+            <strong data-bind="text: language">todo</strong>
+            and
+            <strong data-bind="text: framework">todo</strong>!
+        </p>
+
+        <p>Language: <input data-bind="value: language" /></p>
+        <p>Framework: <input data-bind="value: framework" /></p>
+
+        <script src="./externals/require.js"></script>
```
mhegazy: require apps usually use `data-main`, wondering why you went this route?

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+
+You'll want to bring your TypeScript files together - both the code you'll be writing as well as any necessary typings files.
+
+To do this, you'll need to create a `tsconfig.json` which contains a list of your input files as well as all your compilation settings.
+Simply create a new file in your project root named `tsconfig.json` and fill it with the following contents.
+
+```json
+{
+    "compilerOptions": {
+        "outDir": "./built/",
```
mhegazy: why not just `outFile`?

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+
+# Putting it all together
+
+Just run:
+
+```shell
+tsc
+```
+
+Now open up `index.html` in your favorite browser and everything should be ready to use!
```
mhegazy: you need a server, something like http-server (https://www.npmjs.com/package/http-server) should do.
```cmd
npm install -g http-server
http-server -o .
```

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+npm install -g typescript typings
+```
+
+You obviously know about TypeScript, but you might not know about typings.
+[Typings](https://www.npmjs.com/package/typings) is a package manager for grabbing definition files.
+
+Next, we'll use typings to grab the declaration files for Knockout:
+
+```shell
+typings install --ambient --save knockout
```
mhegazy: samples i have seen sometimes use `--save-dev` to make it easier for users to get where they want to be by just following the commands.

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+Just run:
+
+```shell
+tsc
+```
+
+Now open up `index.html` in your favorite browser and everything should be ready to use!
+You should see a page that says "Hello from TypeScript and Knockout!"
+Below that, you'll also see two input boxes.
+As you modify their contents and switch focus, you'll see that the original message changes.
```
mhegazy: Are we putting this project on the samples repo?

pages/quick-start/knockout.md
DanielRosenwasser: The options are thus:

1. You do what I did here.
2. You have config that needs to be bundled at the top of *every file* (which is not easy here) or you use `--outFile` and bundle it at the top of that.

You ask why I didn't use `--outFile` below, which I'll answer there.

pages/quick-start/knockout.md
DanielRosenwasser: It's a tough call, but I figured that most Knockout apps *don't* bundle everything into a single `.js` file. Modules aren't covered in the documentation until the *very end*.

As you mentioned above, modules are not always intuitive, so keeping the separate-files-per-module abstraction is *extremely* helpful.

I'm planning to write a "Where to go from here" section for each quick-start tutorial. We can suggest these things there.

pages/quick-start/knockout.md
DanielRosenwasser: Why do I need a server? This worked for me.

pages/quick-start/knockout.md
DanielRosenwasser: I'm confused. What about this would make the tutorial easier? Moreover, why would that be correct to do?

pages/quick-start/knockout.md
DanielRosenwasser: Sure, we can. I have a ready project, but I'd prefer to keep it in sync with the guide before adding it.

pages/quick-start/react-webpack.md
```diff
@@ -80,7 +80,7 @@ This will allow you to debug your final output file as if you were debugging you
 Linking TypeScript allows ts-loader to use your global installation of TypeScript instead of needing a separate local copy.
 If you want a local copy, just run `npm install typescript`.
 
-Finally, we'll grab the declaration files for React using the `typings` utility:
+Finally, we'll use typings to grab the declaration files for React and ReactDOM:
```
sandersn: Very minor: It would be nice if 'typings' stood out better, and the program itself is spelled 'Typings' while the excutable is spelled `typings`. Either one is an improvement.

pages/quick-start/knockout.md
mhegazy: consider then moving the config in a script tag in the html file, and using data-main to point to the entry module.

pages/quick-start/knockout.md
mhegazy: it worked cause it is just a static html, but at runtime it is not. you can have path that will not work at runtime.  and in a browser like IE/edge, it will give you a warning about local scripts..

pages/quick-start/knockout.md
mhegazy: it does not make it easier, it just makes it easier to get to your final state. i am assuming you are doing this, test it, and check it in in some source control system, where other ppl can use it as well. So you want it to be duplicate for other users, or for other machines, `npm install` is probably what you need. 
it is up to you, just raised the point.

pages/quick-start/knockout.md
DanielRosenwasser: You can't do that because data-main is loaded asynchronously, and might occur before the config loads.

pages/quick-start/react-webpack.md
DanielRosenwasser: Actually, I don't see where typings is capitalized in their documentation except potentially at the beginning of sentences, so I've been careful here. Can you point out where you see otherwisE?

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+
+To do this, you'll need to create a `tsconfig.json` which contains a list of your input files as well as all your compilation settings.
+Simply create a new file in your project root named `tsconfig.json` and fill it with the following contents.
+
+```json
+{
+    "compilerOptions": {
+        "outDir": "./built/",
+        "sourceMap": true,
+        "noImplicitAny": true,
```
sandersn: I don't think this is needed for a tutorial

sandersn: And maybe some of the other options?

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+Notice there are three script tags.
+First, we're including RequireJS itself.
+Then we're mapping the paths of our external dependencies in `configure.js` so that RequireJS knows where to look for them.
+Finally, we're calling `require` with a list of modules we'd like to load.
+
+# Add a TypeScript configuration file
+
+You'll want to bring your TypeScript files together - both the code you'll be writing as well as any necessary typings files.
```
sandersn: Would it be worthwhile to move this step earlier?

I'm using tide right now, which requires a tsconfig to provide any language service support at all, but I don't if any editor will be able to give good completions from `ko` without tsconfig.

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+{
+    "compilerOptions": {
+        "outDir": "./built/",
+        "sourceMap": true,
+        "noImplicitAny": true,
+        "module": "amd",
+        "target": "es5"
+    },
+    "files": [
+        "./typings/main.d.ts",
```
sandersn: does this list need to be explicit? I would like some explanation the typings path if not -- why `./typings/main.d.ts` instead of the many other .d.ts files in the typings tree?

pages/quick-start/knockout.md
sandersn: I would add "Just type in the boxes and press enter to update the message." (or something like that)

I thought it was broken at first because it was not updating as I typed.

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+You can learn more about `tsconfig.json` files [here](../tsconfig.json.md).
+
+# Putting it all together
+
+Just run:
+
+```shell
+tsc
```
sandersn: I got 

```ts
src/hello.ts(4,15): error TS2503: Cannot find namespace 'ko'.
src/hello.ts(5,16): error TS2503: Cannot find namespace 'ko'.
```

And the language service pointed out that same thing. The resulting code works though.

sandersn: ./src/hello.ts:

```ts
import * as ko from "knockout";

class HelloViewModel {
    language: ko.Observable;
    framework: ko.Observable;

    constructor(language: string, framework: string) {
        this.language = ko.observable(language);
        this.framework = ko.observable(framework);
    }
}

ko.applyBindings(new HelloViewModel("TypeScript", "Knockout"));
```

./src/config.ts:

```ts
declare var require: any;
require.config({
    paths: {
        "knockout": "externals/knockout-3.4.0",
    }
});
```

./tsconfig.json

```json
{
    "compilerOptions": {
        "outDir": "./built/",
        "sourceMap": true,
        "noImplicitAny": true,
        "module": "amd",
        "target": "es5"
    },
    "files": [
        "./typings/main.d.ts",
        "./src/config.ts",
        "./src/hello.ts"
    ]
}
```

pages/quick-start/knockout.md
mhegazy: never mind, somehow i was looking at the npm install line when i made the comment. but even that uses -g which makes the whole comment meaningless. sorry about the noise.

pages/quick-start/knockout.md
DanielRosenwasser: We're trying to put people on the best path to writing code, so we did this for the other quick-start guide.

pages/quick-start/knockout.md
DanielRosenwasser: Yes, I'm planning to do that in the React tutorial as well actually.

pages/quick-start/react-webpack.md
sandersn: I looked at their Readme.md:
1. Title
2. Inline comment
3. Feature List

## (1) Typings

The manager for TypeScript definitions.
Quick Start

....

```shell
# Install (2) Typings CLI utility.
npm install typings --global
```
...

If a project uses (3) Typings, you can install it locally - try `typings install npm:popsicle`

pages/quick-start/knockout.md
sandersn: makes sense

pages/quick-start/react-webpack.md
DanielRosenwasser: Thanks for pointing that out! I was looking at [their npm page](https://www.npmjs.com/package/typings).

blakeembrey: Feel free to ping me for anything you need too :smile: I just started watching the repo. I'm trying to use Typings to describe the tool, `typings` for the CLI and typings (lowercase) to generally describe type definitions and parity with TypeScript typings field in `package.json`.

Things could always use some docs love, and the TypeScript native support is something that should be better described.

DanielRosenwasser: @blakeembrey thanks! I didn't want to cause a lot of noise on your end, but I appreciate that. Please let me know if you see anything that sticks out specifically in these quick-start guides. I'm trying to base these write-ups on my understanding of best practices for using these tools.

DanielRosenwasser:
@sandersn can you take another look? I've addressed most of your feedback.

sandersn:
The changes look good. I still need to test whether the compiler can resolve `ko`. Does changing `configure.ts` to `require-configure.ts` do this? That's the only sources change I saw in the new commits.

pages/quick-start/knockout.md
```diff
@@ -0,0 +1,188 @@
...
+# Write some code
+
+Let's write our first TypeScript file using Knockout.
+First, create a new file in your `src` directory named `hello.ts`.
+
+```ts
+import * as ko from "knockout";
+
+class HelloViewModel {
+    language: ko.Observable
```
sandersn: `ko.Observable` should be `KnockoutObservable<string>`

pages/quick-start/knockout.md
DanielRosenwasser: As mentioned below, I needed `KnockoutObservable` instead of `ko.Observable`

PR #185, opened 2016-03-14
Update modules

### LINKED ISSUE #4: Module transition guide

### LINKED ISSUE #16: Review Modules for language changes
Update this file so that it is up-to-date with regard to new language features, especially around union types and ES6 features

sandersn:
Fixes #16
Fixes #4 since commit 2b3808d

Update modules and namespace documentation.
1. Move Namespace and legacy Module documentation to a new folder named appendices. I'm not sure of the name, but we definitely need a place for documentation that's not part of the handbook. I'm thinking of things like FAQs, documentation on deprecated features, new-feature upgrade guides, etc.
2. Update code samples.
3. Cut out duplication and wordiness.


pages/Modules.md
```diff
@@ -1,11 +1,9 @@
-> **A note about terminology:**
-It's important to note that in TypeScript 1.5, the nomenclature has changed.
-"Internal modules" are now "namespaces".
-"External modules" are now simply "modules", as to align with [ECMAScript 2015](http://www.ecma-international.org/ecma-262/6.0/)'s terminology, (namely that `module X {` is equivalent to the now-preferred `namespace X {`).
-
 # Introduction
 
-Starting with the ECMAScript 2015, JavaScript has a concept of modules. TypeScript shares this concept.
+Starting with the ECMAScript 2015, JavaScript has modules. TypeScript shares this concept.
+Modules are packages of code and declarations that declare their dependencies using `import` and `export`.
```
bowdenk7: This is confusing, could we cut "and declarations":

Modules are packages of code that declare their dependencies...

pages/Modules.md
```diff
@@ -49,6 +50,7 @@ export class ZipCodeValidator implements StringValidator {
 Export statements are handy when exports need to be renamed for consumers, so the above example can be written as:
 
 ```ts
+import { StringValidator } from "Validation";
 class ZipCodeValidator implements StringValidator {
     isAcceptable(s: string) {
         return s.length === 5 && numberRegexp.test(s);
```
bowdenk7: a couple lines below there is: 
```js
export { ZipCodeValidator };
export { ZipCodeValidator as mainValidator };
```
Why do we export both?

pages/Modules.md
```diff
@@ -81,9 +83,9 @@ Optionally, a module can wrap one or more modules and combine all their exports
 ##### AllValidators.ts
 
 ```ts
-export * from "./StringValidator"; // exports interface StringValidator
-export * from "./LettersOnlyValidator"; // exports class LettersOnlyValidator
+export * from "./Validation; // exports interface StringValidator
 export * from "./ZipCodeValidator";  // exports class ZipCodeValidator
+export * from "./ParseIntZipCodeValidator"; // exports class ParseIntZipCodeValidator
 ```
```
bowdenk7: Should we add what the import for this looks like?

```
//someConsumer.ts
import {ZipCodeValidator, ParseIntZipCodeValidator} from "./AllValidators";
```

pages/Modules.md
```diff
@@ -81,9 +83,9 @@ Optionally, a module can wrap one or more modules and combine all their exports
...
 
 # Import
```
bowdenk7: Can we get rid of the `declare` on line ~140? -> `let $: JQuery;`

bowdenk7: It's also kind of unclear that you can 
```js
import <whatever_name_you_want> from "JQuery";
```
in the default export example.

pages/Modules.md
```diff
@@ -212,16 +214,16 @@ console.log(num); // "123"
 
 # `export =` and `import = require()`
```
bowdenk7: I think we should mention somewhere in here that the ES6 style import/export is preferred.

pages/Modules.md
```diff
@@ -389,88 +321,41 @@ validators["Letters only"] = new LettersOnlyValidator();
...
-
-import { ZipCodeValidator as Zip } from "./ZipCodeValidator";
-
-if (needZipValidation) {
-    System.import("./ZipCodeValidator").then((ZipCodeValidator: typeof Zip) => {
-        var x = new ZipCodeValidator();
-        if (x.isAcceptable("...")) { /* ... */ }
-    });
-}
+export declare var x: string;
```
bowdenk7: Where do we mention what `declare` means? We should also mention that it is implicit in .d.ts files.

pages/Modules.md
```diff
@@ -723,34 +544,27 @@ Here is a test for our ProgrammerCalculator class:
 import { Calculator, test } from "./ProgrammerCalculator";
 
 let c = new Calculator(2);
-test(c, "001+010="); // prints 3
+test(c, "001+010="); // should print 3
 ```
 
-## Do not use namespaces in modules
```
bowdenk7: Don't we still not want people to do this? Did this get moved elsewhere?

pages/Modules.md
DanielRosenwasser: It's also probably not a good idea to call these "packages" since those are thing in npm

sandersn: @bowdenk7 It's technically true for TypeScript -- not JavaScript -- but since we're describing ES2015's concept here it **is** pretty confusing. Removed

@DanielRosenwasser bundles? It's hard to find a neutral word here.

sandersn: @RyanCavanaugh suggests 'unit', which sounds great to me.

pages/Modules.md
sandersn: To show both syntaxes for exporting. I'll add a comment to make that clear.

pages/Modules.md
sandersn: It's labelled as a d.ts, so I think the `declare` should stay. But I'll point out that this is the *interface* for JQuery, not JQuery itself.

pages/Modules.md
sandersn: It should be in Variable Declarations or Writing Declaration Files. I need to double-check that both of those files come before this one. They *should* but I don't know.

pages/Modules.md
sandersn: I reworded and added an inline syntax example of default imports.

pages/Modules.md
sandersn: done

pages/Modules.md
sandersn: Added a link to Writing Declaration Files.

pages/Modules.md
sandersn: It is largely redundant with another discussion in Namespaces and Modules.md that's now in appendices/Modules.md. I kept a little bit of this one but mostly went with the one moved from Namespaces and Modules.

sandersn:
@bowdenk7, @DanielRosenwasser can you take another look now that I addressed the PR comments?

pages/Modules.md
```diff
@@ -14,9 +12,11 @@ Modules are declarative; the relationships between modules are specified in term
...
-Well-known modules loaders used in JavaScript are the [CommonJS](https://en.wikipedia.org/wiki/CommonJS) module loader for Node.js and [require.js](http://requirejs.org/) for Web applications.
+Well-known module loaders used in JavaScript are the [CommonJS](https://en.wikipedia.org/wiki/CommonJS) module loader for Node.js and [require.js](http://requirejs.org/) for Web applications.
+TypeScript can generate appropriate code for Node.js ([CommonJS](http://wiki.commonjs.org/wiki/CommonJS)), require.js ([AMD](https://github.com/amdjs/amdjs-api/wiki/AMD)), isomorphic ([UMD](https://github.com/umdjs/umd)), [SystemJS](https://github.com/systemjs/systemjs), or [ECMAScript 2015 native modules](http://www.ecma-international.org/ecma-262/6.0/#sec-modules) (ES6) module-loading systems.
 
-In TypeScript, just as in ECMAScript 2015, any file containing a top-level `import` or `export` is considered a module.
+> **A note about terminology:**
+It's important to note that before TypeScript 1.5, modules were called "external modules" and namespaces were called "internal modules".
```
mhegazy: can we add a link `[namespaces](./Namespaces.md)` ?

sandersn: I moved Namespaces.md into `./appendices/`, but I can definitely link to it there.

sandersn: done

pages/appendices/Namespaces.md
```diff
@@ -1,19 +1,25 @@
-> **A note about terminology:**
```
mhegazy: why did we move this file? namespaces are a language construct and i would expect them to be at the top level

sandersn: I don't think we should encourage namespaces for new code when modules have essentially obsoleted them.

bowdenk7:
:+1: 
Looks good enough to publish.

pages/Modules.md
```diff
@@ -14,9 +12,11 @@ Modules are declarative; the relationships between modules are specified in term
 
 Modules import one another using a module loader.
 At runtime the module loader is responsible for locating and executing all dependencies of a module before executing it.
-Well-known modules loaders used in JavaScript are the [CommonJS](https://en.wikipedia.org/wiki/CommonJS) module loader for Node.js and [require.js](http://requirejs.org/) for Web applications.
+Well-known module loaders used in JavaScript are the [CommonJS](https://en.wikipedia.org/wiki/CommonJS) module loader for Node.js and [require.js](http://requirejs.org/) for Web applications.
```
DanielRosenwasser: CommonJS and AMD

pages/Modules.md
```diff
@@ -14,9 +12,11 @@ Modules are declarative; the relationships between modules are specified in term
...
+TypeScript can generate appropriate code for Node.js ([CommonJS](http://wiki.commonjs.org/wiki/CommonJS)), require.js ([AMD](https://github.com/amdjs/amdjs-api/wiki/AMD)), isomorphic ([UMD](https://github.com/umdjs/umd)), [SystemJS](https://github.com/systemjs/systemjs), or [ECMAScript 2015 native modules](http://www.ecma-international.org/ecma-262/6.0/#sec-modules) (ES6) module-loading systems.
```
DanielRosenwasser: This reads as "or ECMAScript 2015 native modules (ES6) module-loading systems"

pages/Modules.md
```diff
@@ -49,13 +50,14 @@ export class ZipCodeValidator implements StringValidator {
...
 class ZipCodeValidator implements StringValidator {
     isAcceptable(s: string) {
-        return s.length === 5 && numberRegexp.test(s);
+        return numberRegexp.test(s);
     }
 }
-export { ZipCodeValidator };
-export { ZipCodeValidator as mainValidator };
+export { ZipCodeValidator as MainValidator }; // Rename to MainValidator
+export { ZipCodeValidator }; // Also export without rename
```
DanielRosenwasser: renaming

pages/Modules.md
```diff
@@ -127,22 +135,23 @@ import "./my-module.js";
 
 Each module can optionally export a `default` export.
 Default exports are marked with the keyword `default`; and there can only be one `default` export per module.
-`default` exports are imported using a different import form.
+`default` exports are imported using a different import form &mdash; `import <name> from "library";`.
 
 `default` exports are really handy.
-For instance, a library like JQuery might have a default export of `jQuery` or `$`, which we'd probably also import under the name `$` or `jQuery`.
+For instance, a library like JQuery would have a default export of `$`, which we'd import under the name `$`.
```
DanielRosenwasser: I don't know if I agree with this change. The reason it was written like that is because we want to point out that you get the classic jQuery object

pages/Modules.md
DanielRosenwasser: The problem with this is that it kind of glazes over where you would place this file - and now you've mislead users into thinking they can just plop this file anywhere. You can't. It needs to be bundled with the library or submitted to the registry.

pages/Modules.md
```diff
@@ -389,88 +329,42 @@ validators["Letters only"] = new LettersOnlyValidator();
...
 
 ## Ambient Modules
 
-In Node.js, most tasks are accomplished by loading one or more modules.
-We could define each module in its own `.d.ts` file with top-level export declarations, but it's more convenient to write them as one larger `.d.ts` file.
-To do so, we use a construct similar to ambient namespaces, but we use the `module` keyword and the quoted name of the module which will be available to a later import.
-For example:
+Let's look at a convenient technique for writing `.d.ts` files for multi-module libraries.
+We could write one `.d.ts` file for each `.js` file.
+But it's more convenient to cram multiple modules in a single large `.d.ts`.
```
DanielRosenwasser: 1. Join this and the above sentences
2. You jump to this too readily. Remember that ambient module declarations are the devil and only a point of convenience when a better solution doesn't exist. At the very least, don't say that it's more convenient. Say that it can *sometimes* be convenient if declaration files are not available.

pages/Modules.md
```diff
@@ -389,88 +329,42 @@ validators["Letters only"] = new LettersOnlyValidator();
...
+To do so, we use the `module` keyword and the quoted name of the module which would normally be obtained from the module's filename.
```
DanielRosenwasser: module's path

pages/Modules.md
```diff
@@ -492,7 +386,8 @@ declare module "path" {
 }
 ```
 
-Now we can `/// <reference>` `node.d.ts` and then load the modules using `import url = require("url");`.
+Now we `/// <reference path="node.d.ts">`.
```
DanielRosenwasser: Now we can add it to our project's list of included files, or add a `/// <reference path="node.d.ts">` to the top of the file.

pages/Modules.md
```diff
@@ -538,26 +432,24 @@ export default function getThing() { return 'thing'; }
...
 
 ```ts
 export class SomeType { /* ... */ }
 export function someFunc() { /* ... */ }
 ```
 
-Conversly when importing:
-
-### Explicitlly list imported names
+### If you're importing multiple objects, explicitly list imported names
```
DanielRosenwasser: These section names are so weird. I could kind of see it being changed to `Explicitly listing imported names` and then giving the advice.

pages/Modules.md
```diff
@@ -592,83 +484,21 @@ A common JS pattern is to augment the original object with *extensions*, similar
 As we've mentioned before, modules do not *merge* like global namespace objects would.
 The recommended solution is to *not* mutate the original object, but rather export a new entity that provides the new functionality.
 
-Consider a simple calculator implementation defined in module `Calculator.ts`.
+Consider a simple calculator interface defined in module `Calculator.ts`.
```
DanielRosenwasser: We're going to get people complaining about errors on this page of calling that an interface.

pages/Modules.md
```diff
@@ -592,83 +484,21 @@ A common JS pattern is to augment the original object with *extensions*, similar
...
-    for (let i = 0; i < input.length; i++) {
-        c.handelChar(input[i]);
-    }
-
-    console.log(`result of '${input}' is '${c.getResult()}'`);
+export declare class Calculator {
+    // implementation is left as an exercise for the reader
+    protected processDigit(digit: string, currentValue: number);
+    protected processOperator(operator: string);
+    protected evaluateOperator(operator: string, left: number, right: number);
```
DanielRosenwasser: These should be removed because they were just helper functions in the implementation.

pages/Modules.md
```diff
@@ -592,83 +484,21 @@ A common JS pattern is to augment the original object with *extensions*, similar
...
+    public handleChar(char: string);
+    public getResult(): number;
 }
+export declare function test(c: Calculator, input: string);
```
DanielRosenwasser: Newline above.

This is now extremely abstract (no pun intended). At least give some documentation on what this is supposed to do.

DanielRosenwasser: Or actually, shouldn't the implementation of `test` be in `TestCalculator.ts`?

pages/Modules.md
```diff
@@ -723,34 +553,27 @@ Here is a test for our ProgrammerCalculator class:
...
+
+A common mistake is to try to use the `/// <reference ... />` syntax to refer to a module file, rather than using an `import` statement.
+However, module resolution is separate from triple-slash references.
+See [Module Resolution](./Module Resolution.md) to learn how module resolution works.
+See [Triple-Slash References](./Triple-Slash Directives.md) to learn how triple-slash references work.
+
+## Module Bundling
+
+Both JavaScript and TypeScript have a one-to-one correspondence between files and modules.
+One effect of this is that it's not possible to use the `--outFile` compiler switch to concatenate multiple JavaScript module files into a single JavaScript file.
```
DanielRosenwasser: That's not true anymore if you're using System or AMD

pages/appendices/Namespaces.md
mhegazy: We do not encourage them as a mean of organizing your project, but there is nothing wrong with using namespaces within modules. 

As a matter of fact, there are scenarios that require the use of namespaces. e.g. extending the static side of a class.

We should add notes in the file saying do not use these as a way to partition your code, but use modules. but hiding them does not do ppl much good.

pages/Modules.md
sandersn: done

pages/Modules.md
sandersn: changed to "or native ECMAScript 2015 (ES6) module-loading systems."

pages/Modules.md
sandersn: added "or `jQuery`" back in.

pages/Modules.md
sandersn: good point. The whole "cram multiple modules" thing is too subtle a discouragement. I'll warn people away more strongly.

sandersn: done

pages/Modules.md
sandersn: done

pages/Modules.md
sandersn: This section doesn't even really address the problem you raise -- it's just a one-paragraph summary of "Writing Declaration Files". This whole section needs to be expanded to address *actual* module-based d.ts usage.

pages/Modules.md
sandersn: "Our project's list of included files" -- you mean tsconfig's `"files"` member, right?

sandersn: OK, I modified it. Let me know what you think.

pages/Modules.md
sandersn: solution: delete "interface" :)

pages/Modules.md
sandersn: You're right. I decided to change all the headings -- this one becomes "Importing Multiple Objects".

pages/Modules.md
sandersn: Well, `processDigit` is used later in `ProgrammerCalculator`, and I want to give people a skeleton to reimplement `Calculator` if they want. So I think I'll leave these.

pages/Modules.md
sandersn: The example makes more sense if there's two things in each `Calculator` module. I'll just rename the functions to `calculate` or something. (My initial instinct was `run`, but that's pretty vague.)

pages/Modules.md
sandersn: Actually, I think this discussion should be in Module Resolution.md. I'll add a link to that page too.

pages/appendices/Namespaces.md
sandersn: There are extensive examples in the Namespaces file of how to structure your application (basically a dupe of the one in the Modules file). I think we should remove those and replace them with the way we actually expect namespaces to be used. Eg extending the static side of a class.

pages/Modules.md
sandersn: fixed

sandersn:
OK, I think I addressed all the comments from the previous round of reviewing. @DanielRosenwasser @mhegazy can you take another look?

RyanCavanaugh:

PR #186, opened 2016-03-15
This class types

### LINKED ISSUE #62: 'this' typing
We need to cover `this` typing.

sandersn:
Add `this` types to Advanced Types.

Fixes #62

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -317,10 +317,9 @@ On the other hand, if you can't express some shape with an interface and you nee
 
 # Polymorphic `this` types
 
-TypeScript has a `this` type in addition to JavaScript's `this` expression.
-The `this` type in a class or interface represents a type that is a *subtype* of the containing type.
-This makes hierarchical fluent interfaces much easier to express.
-For example, take a simple calculator that returns `this` after each operation:
+A polymorphic `this` type represents a type that is the *subtype* of the containing class or interface.
+This makes hierarchical fluent interfaces much easier to express, for example.
+Take a simple calculator that returns `this` after each operation:
 
 ```ts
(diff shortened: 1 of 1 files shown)


pages/Advanced Types.md
```diff
@@ -313,4 +313,86 @@ As we mentioned, type aliases can act sort of like interfaces; however, there ar
...
+        .sin()
+        .add(1)
+        .currentValue();
+```
+
+Without `this` types, `ScientificCalculator` would not have been able to extend `BasicCalculator` and keep the fluent interface.
+`multiply` would have returned `BasicCalculator`, which doesn't have the `sin` method.
+However, with `this` types, `multiply` returns `this`, which is `ScientificCalculator` here.
+
+You can also use `this` types to describe mixin patterns:
```
mhegazy: not sure this is accurate.. consider:

```ts
class A {
    a(): this { return this; }
}

class B {
    b(): this { return this; }
    extend<T>(other: T): this & T { return <any>this; }
}


var b = new B();

var a = new A();

var extended = b.extend(a);

extended.a().b() // Error: b does not exist on type A
```

this types, the way we implement them, do not play well with intersection types in mixin patterns.

sandersn: Should I drop this whole mixin section? I wrote it up based on a small sample in the What's New for 1.7, but it seemed pretty undeveloped.

mhegazy: i agree. it looks like it works, untill you dot into it, and then it falls off. we should not be leading ppl there intentionally.

sandersn: Removed

mhegazy:
:+1:

PR #192, opened 2016-03-16
Document specialized overload bubbling behavior.

DanielRosenwasser:


pages/Declaration Merging.md
```diff
@@ -55,27 +55,59 @@ Of note, too, is that in the case of interface `A` merging with later interface
...
+    clone(animal: Cat): Cat;
+    clone(animal: Sheep): Sheep;
+    clone(animal: Animal): Animal;
+}
+```
+
+Notice that the elements of each group maintains the same order, but the groups themselves are merged with later overload sets ordered first.
+
+One exception to this rule is specialized signatures.
+If a signature has a parameter whose type is a *single* string literal type (e.g. not a union of string literals), then it will be bubbled toward the top of its overload list.
```
sandersn: add 'merged': "its **merged** overload list"

sandersn:
:+1:

PR #198, opened 2016-03-21
Add discussion of destructuring.

### LINKED ISSUE #31: Destructuring
Declarations with destructuring, assignments, and samples.

sandersn:
Fixes #31.

### DIFF
--- a/pages/Variable Declarations.md
+++ b/pages/Variable Declarations.md
@@ -539,15 +539,12 @@ let {a, b}: {a: string, b: number} = o;
 Default values let you specify a default value in case a property is undefined:
 
 ```ts
-o = {a: "foo", c: "bar"};
-let {a, b: newName = 1001}: {a: string, b?: number} = o;
+function keepWholeObject(wholeObject: {a: string, b?: number}) {
+    let {a, b = 1001} = wholeObject;
+}
 ```
 
-Note that in this example, the type annotation is required because otherwise the compiler will not expect the `b` property to come from the object `o`.
-
-Use destructuring with care.
-Even though you can nest destructuring, deeply nested expressions get *really* hard to understand even without piling on renaming, default values, and type annotations.
-Try to keep destructuring assignments small and simple.
+`keepWholeObject` now has a variable for `wholeObject` as well as the properties `a` and `b`, even if `b` is undefined.
 
 ## Function declarations
 
@@ -565,19 +562,26 @@ But specifying defaults is more common for parameters, and getting defaults righ
 First of all, you need to remember to put the type before the default value.
 
 ```ts
-function f({a, b}: C = {a: "", b: 0}): void {
+function f({a, b} = {a: "", b: 0}): void {
     // ...
 }
 f(); // ok, default to {a: "", b: 0}
 ```
 
-Then, you need to remember to give a default for optional properties on the property instead of the main initializer.
+Then, you need to remember to give a default for optional properties on the destructured property instead of the main initializer.
 Remember that `C` was defined with `b` optional:
 
 ```ts
-function f({a, b = 0}: C = {a: ""}): void {
+function f({a, b = 0} = {a: ""}): void {
     // ...
 }
 f({a: "yes"}) // ok, default b = 0
 f() // ok, default to {a: ""}, which then defaults b = 0
+f({}) // error, 'a' is required if you supply an argument
 ```
+
+Use destructuring with care.
+As the previous example demonstrates, anything but the simplest destructuring expressions have a lot of corner cases.
+This is especially true with deeply nested destructuring, which gets *really* hard to understand even without piling on renaming, default values, and type annotations.
+Try to keep destructuring expressions small and simple.
+You can always write the assignments that destructuring would generate yourself.
--- a/pages/Variable Declarations.md
+++ b/pages/Variable Declarations.md
@@ -440,7 +440,7 @@ In this section, we'll give a short overview.
 
 ## Array destructuring
 
-The simplest form of destructuring is destructuring array assignment:
+The simplest form of destructuring is array destructuring assignment:
 
 ```ts
 let input = [1, 2];
@@ -449,8 +449,8 @@ console.log(first); // outputs 1
 console.log(second); // outputs 2
 ```
 
-This binds two new variables named `first` and `second`.
-This is equivalent to using indexing, but much more convenient:
+This creates two new variables named `first` and `second`.
+This is equivalent to using indexing, but is much more convenient:
 
 ```ts
 first = input[0];
@@ -473,7 +473,7 @@ function f([first, second]: [number, number]) {
 f(input);
 ```
 
-You can bind a variable to remaining items in a list using the syntax `...name`:
+You can create a variable for the remaining items in a list using the syntax `...name`:
 
 ```ts
 let [first, ...rest] = [1, 2, 3, 4];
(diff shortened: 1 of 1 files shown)


pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+    b: 12,
+    c: "bar"
+}
+let { a: newName1, b: newName2 } = o;
+```
+
+This is even more convenient than array destructuring, but the directionality is confusing.
+For each pair, the identifier on the left side of the colon is the property name from the object, and the identifier on the right side of the colon is the newly bound variable.
+The direction is left-to-right:
+For example, `{ a: newName1 }` means bind the property 'a' to the variable 'newName1'.
```
mhegazy: "bind" is a bit too technical. i would say something like:
declare a new variable `newName1` and initialize it to the value of `o.a`

mhegazy: One thing that ppl usually find confusing is the use of `:` to rename and how to declare the type annotation for the new name. we should have a paragraph for this, either here or earlier to cover both array and object destructuring variable declarations.

mhegazy: Just a section to say, please note that colon here does not mean a type, but rather a new name, and the type has to be  following the whole binding pattern, and you will need to repeat the names of the properties you defined.

mhegazy: it might also be useful to say `a: b` reads as `a as b`.

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+    // code goes here ..
+}
+
+// Fix 2: Explicit code
+for (let { b } of objects) {
+    if (b && b.length > 0) {
+        let first = b[0];
+        // code goes here ...
+    }
+}
```
mhegazy: we should add another section about function parameters, specially the diffrent ways you can initialize a paramter, and what that means:
```ts
function foo({a, b} = {});
function foo({a = 0, b = 1});
function foo({a = 0, b = 1} = {});
function foo({a, b} = {a: 0, b: 1});
```

see https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#improved-checking-for-destructuring-object-literal for more details.

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+
+# Destructuring
+
+Another ECMAScript 2015 feature that TypeScript has is destructuring.
+For a complete reference, see [the article on the Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).
+In this section, we'll give a short overview.
+
+## Array destructuring
+
+The simplest form of destructuring is destructuring array assignment:
```
DanielRosenwasser: is an array destructuring assignment.

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+
+```ts
+let input = [1, 2];
+let [first, second] = input;
+console.log(first); // outputs 1
+console.log(second); // outputs 2
+```
+
+This binds two new variables named `first` and `second`.
```
DanielRosenwasser: `/binds/creates/`

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+This is equivalent to using indexing, but much more convenient:
```
DanielRosenwasser: but is

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+```ts
+let o = {
+    a: "foo",
+    b: 12,
+    c: "bar"
+}
+let { a: newName1, b: newName2 } = o;
+```
+
+This is even more convenient than array destructuring, but the directionality is confusing.
```
DanielRosenwasser: Why is this more convenient that array destructuring? Technically array destructuring is more convenient because users don't have to write

```ts
let { 0: x, 1: y } = [1,2,3,4];
```

Also, change to "can be confusing"

pages/Variable Declarations.md
DanielRosenwasser: Yes, you can omit these three sentences if you just write

> You can read each of these `a: b` pairs as "a *as* b".
> The idea is that `let { a as newName1 } = o` is basically the same as `let newName = o.a`.

pages/Variable Declarations.md
sandersn: How about:

"You'll probably use this even more than array destructuring, but the syntax can be confusing."

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+For each pair, the identifier on the left side of the colon is the property name from the object, and the identifier on the right side of the colon is the newly bound variable.
+The direction is left-to-right:
+For example, `{ a: newName1 }` means bind the property 'a' to the variable 'newName1'.
+
+This direction is important to remember because of two additional features: shortcut names and default values.
```
DanielRosenwasser: I think most destructuring cases will us "shortcut names". Since that's the common case, consider making that the first example you demonstrate with object destructuring.

That way you don't have to create a new term for this feature.

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+
+All the above features can be nested and used anywhere JavaScript assigns values to variables, like `for` loops and function calls.
+Here's a more complex example:
+
+```ts
+let objects: { a: number, b?: number[] }[] = [
+  { "a": 12, "b": [3, 4] },
+  { "a": 21 },
+  { "a": 11, "b": [4, 3] }];
+for (let { b: [first] = [0] } of objects) {
```
DanielRosenwasser: Add a newline above

pages/Variable Declarations.md
sandersn: Yeah, I started the other way because it makes complex cases easier to read. But I don't think we want to encourage any complex uses of this feature at all. I'll flip the order.

pages/Variable Declarations.md
```diff
@@ -430,4 +430,150 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+    // code goes here ...
+}
+```
```
DanielRosenwasser: Part of me isn't so sure this example is super useful. I get what it's going for, but let's not encourage people to do this.

Almost every time I've ever started writing a nested destructuring, I stopped myself because it was going to be an unreadable mess.

pages/Variable Declarations.md
sandersn: done

pages/Variable Declarations.md
sandersn: done (except 'an' -- this is a general statement and the indefinite article is optional there.)

pages/Variable Declarations.md
sandersn: done

pages/Variable Declarations.md
sandersn: Never mind, all this text went away when I reordered the discussion.

pages/Variable Declarations.md
sandersn: I think I'll cut the complex example and put a stronger warning not to be foolhardy in its place.

pages/Variable Declarations.md
DanielRosenwasser: Thinking about it more, I think we should steer away from saying that this is confusing.

pages/Variable Declarations.md
DanielRosenwasser: That sounds like a good idea

pages/Variable Declarations.md
sandersn: done

sandersn:
I added a commit to address the PR comments.

pages/Variable Declarations.md
```diff
@@ -504,76 +504,80 @@ let o = {
...
 
-Here the directionality of the syntax becomes quite confusing.
-In `b: newName2 = 1001`, the new variable is `newName2`, and its value is `o.b` unless `o.b` is undefined, in which case it is `1001`.
-Use default values with care.
+Confusingly, the colon here does *not* indicate the type.
+The type, if you specify it, still needs to be specified after the entire destructuring:
 
-## Complex example
+```ts
+let {a, b}: {a: string, b: number} = o;
```
mhegazy: that is kind of meaningless, as `o` probably has a type, and you really do not need to do this. a better example would be one in function parameter position where you need to define the type:

```ts
function test({a, b}: {a:string, b: number}) { ... }
```

pages/Variable Declarations.md
```diff
@@ -504,76 +504,80 @@ let o = {
...
-let objects: { a: number, b?: number[] }[] = [
-  { "a": 12, "b": [3, 4] },
-  { "a": 21 },
-  { "a": 11, "b": [4, 3] }];
-for (let { b: [first] = [0] } of objects) {
-    // code goes here ...
-}
+o = {a: "foo", c: "bar"};
+let {a, b: newName = 1001}: {a: string, b?: number} = o;
 ```
```
mhegazy: i find this confusing. so you know that `o` does not have a `b`, so it is always going to be `1001`, so why put it here?

pages/Variable Declarations.md
```diff
@@ -504,76 +504,80 @@ let o = {
...
+First of all, you need to remember to put the type before the default value.
 
 ```ts
-// Fix 1: Redundant defaults
-for (let { b: [first = 0] = [0] } of objects) {
-    // code goes here ..
+function f({a, b}: C = {a: "", b: 0}): void {
+    // ...
 }
+f(); // ok, default to {a: "", b: 0}
```
mhegazy: we should note that `f({})` means that `a` and `b` will be undefined.

mhegazy: do not think we should tell ppl to put a type annotation here, unless they want to. if you have defaults, you should not put type annotations, like it is inferred that a is a string, and b is a number, `C` is superfluous here.

pages/Variable Declarations.md
sandersn: I think it's fine for a one-liner introduction of the syntax -- not all examples need to be perfectly realistic. And I want to keep the introduction of destructuring in functions until last.

pages/Variable Declarations.md
sandersn: good point, I did not realise that our inference was that good.

pages/Variable Declarations.md
sandersn: done. I wrote a better example.

mhegazy:
:+1:

sandersn:
@DanielRosenwasser can you take another look?

pages/Variable Declarations.md
```diff
@@ -430,4 +430,158 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+
+```ts
+first = input[0];
+second = input[1];
+```
+
+Destructuring works with already-declared variables as well:
+
+```ts
+[first, second] = input;
```
DanielRosenwasser: One thing that people may run into is object binding patterns, where you have to wrap the expression in parens. Might want to mention that below. Or potentially don't even mentioned this example.

sandersn: I improved the example to show that you can swap variables using assignment. And I want to discuss object destructuring assignment because the syntax is hard to figure out on your own.

pages/Variable Declarations.md
```diff
@@ -460,7 +460,8 @@ second = input[1];
 Destructuring works with already-declared variables as well:
 
 ```ts
-[first, second] = input;
+// swap variables
+[first, second] = [second, input];
```
mhegazy: `[first, second] = [second, first];`

DanielRosenwasser:
Can you sync with the master branch?

sandersn:
Sure, merged.

pages/Variable Declarations.md
```diff
@@ -430,4 +430,167 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+This creates new variables `a` and `b` from `o.a` and `o.b`.
+Notice that you can skip `c` if you don't need it.
+
+Like array destructuring, you can have assignment without declaration:
+
+```ts
+({a, b} = {a: "baz", b: 101});
+```
+
+However, because object destructuring syntax is ambiguous with the start of a block, you have to surround the whole statement with parentheses.
```
DanielRosenwasser: I'd phrase it as

> Notice that we had to surround this assignment with parentheses
This has to do with technical reasons over the way JavaScript is parsed, since an open curly brace usually signals the start of a block.

pages/Variable Declarations.md
```diff
@@ -430,4 +430,167 @@ Using `const` also makes code more predictable when reasoning about flow of data
...
+You can read `a: newName1` as "`a` as `newName1`".
+The direction is left-to-right, as if you had written:
+
+```ts
+let newName1 = o.a;
+let newName2 = o.b;
+```
+
+Confusingly, the colon here does *not* indicate the type.
+The type, if you specify it, still needs to be specified after the entire destructuring:
```
DanielRosenwasser: needs to be written

DanielRosenwasser:
Apart from my current feedback, :+1:.

pages/Variable Declarations.md
sandersn: done

PR #207, opened 2016-03-25
fixed verb

onsentamago:
Fixes #

fix verb,didn't have pass -> didn't pass

### DIFF
--- a/pages/Generics.md
+++ b/pages/Generics.md
@@ -63,7 +63,7 @@ The second way is also perhaps the most common. Here we use *type argument infer
 let output = identity("myString");  // type of output will be 'string'
 ```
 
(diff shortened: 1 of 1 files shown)


pages/Generics.md
```diff
@@ -63,7 +63,7 @@ The second way is also perhaps the most common. Here we use *type argument infer
 let output = identity("myString");  // type of output will be 'string'
 ```
 
-Notice that we didn't have explicitly pass the type in the angle brackets (`<>`), the compiler just looked at the value `"myString"`, and set `T` to its type.
+Notice that we didn't explicitly pass the type in the angle brackets (`<>`), the compiler just looked at the value `"myString"`, and set `T` to its type.
```
mhegazy: i think this was intended to be `have to`

DanielRosenwasser: Yup, Mohamed is correct.

PR #211, opened 2016-03-26
Suggest index signatures for excess properties/update index signatures.

### LINKED ISSUE #156: Suggest an index signature in the excess property checking section
Adapt from https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#strict-object-literal-assignment-checking

DanielRosenwasser:
Fixes #156


pages/Interfaces.md
```diff
@@ -218,10 +221,11 @@ mySearch = function(src, sub) {
 }
 ```
 
-# Array Types
+# Indexable Types
 
-Similarly to how we can use interfaces to describe function types, we can also describe array types.
-Array types have an `index` type that describes the types allowed to index the object, along with the corresponding return type for accessing the index.
+Similarly to how we can use interfaces to describe function types, we can also describe types that we treat a bit like arrays.
```
sandersn: "treat a bit like arrays" -> "index like arrays"

pages/Interfaces.md
```diff
@@ -149,6 +149,9 @@ let mySquare = createSquare(squareOptions);
 
 Since `squareOptions` won't undergo excess property checks, the compiler won't give you an error.
 
+One final way to get around the check is to add an index signature if you are sure that the object will *always* have properties with unknown names.
+We'll discuss index signatures in a bit.
```
mhegazy: i would add an example here using indexers, an say we will discuss them in details later.

pages/Interfaces.md
```diff
@@ -218,10 +221,11 @@ mySearch = function(src, sub) {
 }
 ```
 
-# Array Types
+# Indexable Types
 
-Similarly to how we can use interfaces to describe function types, we can also describe array types.
-Array types have an `index` type that describes the types allowed to index the object, along with the corresponding return type for accessing the index.
+Similarly to how we can use interfaces to describe function types, we can also describe types that we treat a bit like arrays.
+Indexable types have a signature called an *index signature* that describes the types we can use to index into the object, along with the corresponding return type for when we do perform indexing accesses.
```
sandersn: 1. "have a signature called an *index signature*" -> "have an *index signature*"
2. "for when we do perform indexing accesses" -> "for when we index into it"

pages/Interfaces.md
```diff
@@ -230,14 +234,19 @@ interface StringArray {
...
+
+let myStr: string = myArray[0];
 ```
 
-There are two types of supported index types: string and number.
+Above, we have a `StringArray` interface that defines an index signature.
+This index signature states that when a `StringArray` is indexed with a `number`, it will return a `string`.
+
+There are two types of supported index signatures: string and number.
 It is possible to support both types of indexers, with the restriction that the type returned from the numeric indexer must be a subtype of the type returned from the string indexer.
```
sandersn: "with the restriction that" -> "but"

sandersn:
:+1:

DanielRosenwasser:
@sandersn I added a few more changes, care to take a quick look?

pages/Interfaces.md
```diff
@@ -223,8 +233,8 @@ mySearch = function(src, sub) {
 
 # Indexable Types
 
-Similarly to how we can use interfaces to describe function types, we can also describe types that we treat a bit like arrays.
-Indexable types have a signature called an *index signature* that describes the types we can use to index into the object, along with the corresponding return type for when we do perform indexing accesses.
+Similarly to how we can use interfaces to describe function types, we can also describe types that we can "index into" like `a[10]`, or `ageMap["daniel"]`.
+Indexable types have something called an *index signature* that describes the types we can use to index into the object, along with the corresponding return types when indexing.
```
sandersn: get rid of "have something called". The italics already indicate that you're introducing a new term.

pages/Interfaces.md
```diff
@@ -242,7 +252,24 @@ Above, we have a `StringArray` interface that defines an index signature.
 This index signature states that when a `StringArray` is indexed with a `number`, it will return a `string`.
 
 There are two types of supported index signatures: string and number.
-It is possible to support both types of indexers, with the restriction that the type returned from the numeric indexer must be a subtype of the type returned from the string indexer.
+It is possible to support both types of indexers, but the type returned from a numeric indexer must be a subtype of the type returned from the string indexer.
+This is because when indexing with a `number`, JavaScript will actually convert that to a `string` before indexing into an object.
+That means that indexing with `100` (a `number`) is the same thing as indexing with `"100"` (a `string`), so the two need to be consistent.
```
sandersn: :+1: good explanation

pages/Interfaces.md
```diff
@@ -149,8 +149,18 @@ let mySquare = createSquare(squareOptions);
...
+
+```ts
+interface SquareConfig {
+    color?: string;
+    width?: number;
+    [propName: string]: any;
+}
+```
+
+We'll discuss index signatures in a bit, but here we're saying a `SquareConfig` can have any number of properties of any types.
```
sandersn: This is better than the re-assignment workaround. It's *probably* the right fix if you really want no excess property checking, so I'd put it before the two workarounds. Maybe even skip the workarounds?

sandersn:
:+1: again

mhegazy:
:+1:

PR #214, opened 2016-03-28
Move sentence and explain constructor rules better in Classes.md

pdfernhout:
The text did not make sense with the sentence about calling super() between a sentence that leads into an example (with a colon) and the example. 

This issue was introduced in this commit: https://github.com/Microsoft/TypeScript-Handbook/pull/81/commits/d0fa4643d39af583045da6752c4dab34812efcc3

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -73,11 +73,11 @@ tom.move(34);
 This example covers quite a few of the inheritance features in TypeScript that are common to other languages.
 Here we see the `extends` keywords used to create a subclass. You can see this where `Horse` and `Snake` subclass the base class `Animal` and gain access to its features.
 
+Derived classes that contain constructor functions must call `super()` which will execute the constructor function on the base class.
+
 The example also shows how to override methods in the base class with methods that are specialized for the subclass.
 Here both `Snake` and `Horse` create a `move` method that overrides the `move` from `Animal`, giving it functionality specific to each class.
 Note that even though `tom` is declared as an `Animal`, since its value is a `Horse`, when `tom.move(34)` calls the overriding method in `Horse`:
-
-Derived classes that contain constructor functions must call `super()` which will execute the constructor function on the base class.
 
 ```Text
 Slithering...


pages/Classes.md
```diff
@@ -73,12 +73,12 @@ tom.move(34);
 This example covers quite a few of the inheritance features in TypeScript that are common to other languages.
 Here we see the `extends` keywords used to create a subclass. You can see this where `Horse` and `Snake` subclass the base class `Animal` and gain access to its features.
 
+Derived classes that contain constructor functions must call `super()` which will execute the constructor function on the base class.
```
sandersn: Can you make this part of the preceding paragraph? I would add two sentences after this one, something like:

> You have to call `super()` before accessing `this` in the derived constructor.
> Both `Horse` and `Snake` technically do this since neither one actually references `this`.

pdfernhout: @sandersn I have added those two sentences. Since you seemed to want to clarify restrictions about constructors, I also added other sentences to explain another restriction [from the TypeScript spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#8.3.2) about when the 'super' call had to be the first statement in the constructor. 

I moved all that into its own paragraph since it got a bit long. It might be too long now, so feel free to suggest further improvements.

I also added a blank line to split up the earlier paragraph, since the sentence about "This example covers quite a few of the inheritance features..." now leads into the three paragraphs that follow which each explain a different concern (extends, constructor restrictions, and overriding).

I left the changes as two separate commits -- the original one to fix an issue and the second one as an enhancement. Let me know if you would prefer them squashed together into one commit though.

pages/Classes.md
```diff
@@ -71,14 +71,15 @@ tom.move(34);
...
 This example covers quite a few of the inheritance features in TypeScript that are common to other languages.
+
 Here we see the `extends` keywords used to create a subclass. You can see this where `Horse` and `Snake` subclass the base class `Animal` and gain access to its features.
 
+TypeScript has some restrictions about constructors intended to help ensure instances of a class have their properties correctly initialized before the properties are used. Derived classes that contain constructor functions must call `super()` which will execute the constructor function on the base class. You have to call `super()` before accessing `this` in the derived constructor. Both `Horse` and `Snake` technically do that since neither one actually references `this`. A 'super' call must also be the first statement in the constructor of a derived class when the derived class contains initialized properties or has parameter properties. Neither `Horse` or `Snake`, however, have those kind of properties right now. So, as long as those two classes stay that way, they could have console logging or other statements before the `super` call in their constructors without generating compile-time errors.
```
DanielRosenwasser: One sentence per line please.

DanielRosenwasser: Use backticks around "super" in "A 'super' call"

pdfernhout: Thanks for the feedback. I made those improvements and also added parentheses for each super call reference to be consistent.

sandersn:
:+1: I like the changes. We don't care much about commits, so just leave them as-is.

pages/Classes.md
```diff
@@ -71,14 +71,21 @@ tom.move(34);
 ```
 
 This example covers quite a few of the inheritance features in TypeScript that are common to other languages.
+
 Here we see the `extends` keywords used to create a subclass. You can see this where `Horse` and `Snake` subclass the base class `Animal` and gain access to its features.
 
+TypeScript has some restrictions about constructors intended to help ensure instances of a class have their properties correctly initialized before the properties are used.
```
mhegazy: it is not really a TypeScript restriction. it is a JS one. and enforced by the fact that the super, and the this share the same object, and that the object is not well formed, until super call has returned.

mhegazy: how about?
```
Initialization of a Base class properties is done as part of calling the Base constructor in a derived class, i.e. `super()` call.
As a result, `this` can not be used in a constructor of a derived class before `super()` call.
A `super()` call must also be the first statement in the constructor of a derived class when the derived class contains initialized properties or has parameter properties.
```

DanielRosenwasser: @mhegazy's suggestion looks good, with the following applied:

Base -> base
can not -> cannot
before **a** `super()` call

pdfernhout: To clarify, is that change intended to replace just that sentence or all the text in the paragraph? If it is just to change the first sentence, then it introduces redundancy that should at least be removed from the rest. But if is to replace everything, it loses the specificity to the Animal example.

The purpose of the introductory sentence was to motivate the reader to understand the intent behind these rules. I'd agree it could be worded better. 

There is also perhaps a subtle distinction here between different sorts of errors -- compile time, run time, and conceptual/algorithmic. Not to be too pendantic, but [here is a JSFiddle](https://jsfiddle.net/770L012z/ ) with the code that the TypeScript playground output for the Animal example modified with a console.log statement added accessing this (which returns undefined) before a super call in a constructor. So, you can access a not-well-formed object in JavaScript. It is just usually not worth doing. So, in that sense, the TypeScript compiler is attempting to protect against a conceptual/algorithmic mistake by these rules. Now, it is possible this will also mess up a JavaScript virtual machine's optimization processes(?), but that is not quite the same as a language restriction.

And they are slightly overzealous rules as discussed in issue #945 -- because there is no real reason a logging statement of "got here" is going to hurt anything if you do it before a super call in a constructor in a derived class. The issue is more that the compiler can't easily determine how safe such calls are. Java has restrictive rules like this with perhaps more justification given its complex memory model and multithreading, and it is often frustrating that you can't call other functions to prepare data to pass to a superclass.

mhegazy: It is a runtime error to access this before calling super in es6. 

I think we could relax the rule that super has to be the first statement now, that we have the other rule.

pdfernhout: @mhegazy Good point. I was thinking in terms of es5.

pdfernhout: As a point of comparison, [the online Babel version](https://babeljs.io/repl/#?evaluate=true&presets=es2015%2Cstage-3&experimental=false&loose=false&spec=false) lets me put a logging statement before a super -- but only as long as I don't reference this (which generates a compile-time error of `repl: 'this' is not allowed before super()`). 

I also tried the same code directly with Firefox 45.0.1 which supports ES6 classes. While I can add a simple log statement before the super call, I got a runtime error when referencing this before the super call just as @mhegazy said would happen. Here is that error:
<pre>
ReferenceError: |this| used uninitialized in Snake class constructor
</pre>

Here was the code I tried (uncomment selected parts to get various errors -- or to make it compile under TypeScript for the first commented code):
<pre>
class Animal {
    // Uncomment the next line for TypeScript but not for ES6 or Babel
    // name: string;
    constructor(theName) { this.name = theName; }
    move(distanceInMeters) {
        console.log(this.name + ' moved ' + distanceInMeters + 'm');
    }
}

class Snake extends Animal {
    // Uncomment the next line to get a compile-time error in TypeScript
    // locomotion: string = "Slithering";
    constructor(name) {
      console.log("Starting constructor");
      // Uncomment the next line to produce a runtime error for ES6 or a compile-time error for TypeScript or Babel
      // console.log("Is this a runtime error under ES6 to reference this.name too soon?", this.name);
      super(name);
    }
    move(distanceInMeters) {
        console.log("Slithering...");
        super.move(distanceInMeters);
    }
}

let sam = new Snake("Sammy the Python");
sam.move(5);
</pre>

As a meta comment, this pull request is now suffering from a massive case of scope creep. :-) The original pull was just to fix an obvious typo of a misplaced sentence. Then we got into adding sentences. Now we are talking about maybe changing the TypeScript compiler spec and implementation. :-)

Maybe I should revert the pull request back to the original commit and change the title back? Perhaps we could then create a separate issue to discuss additional wording for what the compiler does now and/or perhaps reopen discussion of a spec change in issue #945 (or create a new issue instead of reopen that)? Enhancements are certainly more debatable than fixing a misplaced sentence, and meanwhile users of the Handbook have to navigate around that typo.

mhegazy: This PR adds more explanation, so we should keep it but update the language.
I would say file a new bug for the this restriction.

pdfernhout: To make progress on updating the text, I need a clarifying answer [to the question I asked](https://github.com/Microsoft/TypeScript-Handbook/pull/214#discussion_r57978957) about whether that change you proposed is intended to replace just that sentence or all the text in the paragraph, given it introduces either redundancy or makes the explanation non-specific to the example? The text you outlined is reasonable standing alone. The issue is fitting it in with the next paragraph in the changed text in the pull request which covers much the same ground, and also whether to use that new text you propose to replace an explanation of **why** that restriction is in place in less technical terms (a handbook not being the spec itself).

BTW, to help explore that restriction, here is an improvement to the previous with more commented knobs to tweak. The point of the additional commented options was to explore what happens when you try to access `this` in another function outside the constructor before calling `super()`. Under Chrome 49.0.2623.95 ([which is claimed support ES6 classes either in strict mode or not](https://www.chromestatus.com/feature/4633745457938432)), it seems like the runtime behavior differs on whether in strict mode or not. Also, oddly, the behavior of "this" seems to depend on how you reference it while logging, showing up as the window sometimes, implicitly undefined as others, and with properties that vanish into emptyness when logged. Not sure if that is a Chrome ES6 bug or not. Also, unlike Firefox, the error in the constructor method for referencing `this` is not about calling `super()` but simply: `Uncaught ReferenceError: this is not defined`. So, there are apparently still some bleeding edge browser inconsistencies here... But it does show that under ES6, whether a reference to `this` produces a runtime error during the construction process of an instance including function calls (not just the constructor method itself) depends on the exact situation in practice at the moment (maybe that will change). Circling back to the issue of documentation, then the question is what to tell the TypeScript Handbook reader about the intent of the TypeScript compiler restrictions in constructors (if anything)?
<pre>
// Behavior differs in Chrome 49.0.2623.95 on whether in strict mode or not
// 'use strict';

class Animal {
    // Uncomment the next line for TypeScript but not for ES6 or Babel
    // name: string;
    
    constructor(theName) { this.name = theName; }
    
    move(distanceInMeters) {
        console.log(this.name + ' moved ' + distanceInMeters + 'm');
    }
}

function logThisName(passthru) {
    // the reference to this.name is wrapped in quotes because under Chrome 49 non-strict it can appear as a blank string
    console.log('logThisName start');
    console.log('What is this now', '"' + this + '"');
    // The TypeScript compiler flags the next line with a warning that the field does not exist on Window
    console.log('Example undefined reference using window._aMissingProperty', '"' + window._aMissingProperty + '"');
    console.log('Is it a runtime error under ES6 to reference this.name indirectly before super() call?', '"' + this.name + '"');
    console.log('Is it a runtime error under ES6 to reference this["name"] indirectly before super() call?', '"' + this['name'] + '"');
    console.log('logThisName end');
    return passthru;
}

class Snake extends Animal {
    // Uncomment the next line to get a compile-time error in TypeScript
    // locomotion: string = "Slithering";
    
    constructor(name) {
        console.log("Starting constructor");
      
        // Uncomment the next line to produce a runtime error for ES6 or a compile-time error for TypeScript or Babel
        // console.log("Is it a runtime error under ES6 to reference this.name too soon?", this.name);
    
        // Uncomment the next line to try to fake out the compiler and/or runtime
        // logThisName(null);
      
        // Uncomment the next line and comment the line below it to try to bypass compile-time testructions on calls before super
        // super(logThisName(name));
        super(name);
    }
    
    move(distanceInMeters) {
        console.log("Slithering...");
        super.move(distanceInMeters);
    }
}

let sam = new Snake("Sammy the Python");
sam.move(5);
</pre>

pdfernhout: Out of curiosity, I spent a bit of time looking at the ES6 spec, but did not find the exact place where is says what should happen if you refer to this before calling super in a constructor (either in the constructor method itself or in a function called from the constructor method). It would have something to do with EnvironmentRecord? I looked in some detail at these two sections but did not see where they prohibited using this before super (even as it may not make sense in practice):
[14.5.1 Static Semantics: Early Errors](http://www.ecma-international.org/ecma-262/6.0/#sec-class-definitions-static-semantics-early-errors)
[14.5.14 Runtime Semantics: ClassDefinitionEvaluation](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-classdefinitionevaluation)

Maybe someone could help em out here in finding the exact section of the ES6 spec to understand this better so we could be sure to use precise wording for the Handbook on this?

Otherwise it would be immense scope creep to talk about revising the ES6 Spec to be clearer about something. :-)

pdfernhout: In the interests of fixing the error from a misplaced sentence in the Handbook, what can be done to move this pull request forward? While a change to the pull request has been proposed by @mhegazy which makes sense by itself, it does not fit with the rest of the pull request. So I am not sure what changes to make at this point.

I also requested that [TypeScript issue 945](https://github.com/Microsoft/TypeScript/issues/945) be reopened.

mhegazy: > I also requested that TypeScript issue 945 be reopened.

do you mind filing a new issue

mhegazy: I am sorry i have lost track of this PR. so if all what is needed is moving the sentence around, why not just do that and do not change the text?

pdfernhout: @mhegazy The original pull request just moved that sentence. Then there was a request for more changes by @sandersn which I made, with some other additions I added. In hindsight, it probably would have been best to have suggested back then that we make a new issue for discussion of any additional changes to the text so the original pull request could go through.

pdfernhout: @mhegazy As you requested, I created a new TypeScript issue on changing constructor restrictions ([TypeScript issue #8277](https://github.com/Microsoft/TypeScript/issues/8277)).

mhegazy: i would go back to moving the sentence then.

pdfernhout: @mhegazy To make incremental progress, as suggested, I have brought the pull request back to its original form that just moves this sentence:

> +-Derived classes that contain constructor functions must call `super()` which will execute the constructor function on the base class.

Because the latest push no longer has these changes on GitHub, below are the other changes in the second commit for future reference.

This was my addition (where we have debated whether the origin of the restrictions is more due to JavaScript/ES6 or TypeScript):

> +TypeScript has some restrictions about constructors intended to help ensure instances of a class have their properties correctly initialized before the properties are used.

These were additions by @sandersn (which would always be accurate):

> +You have to call `super()` before accessing `this` in the derived constructor.
> +Both `Horse` and `Snake` technically do that since neither one actually references `this`.

This was my further additions (which would be obsolete if TypeScript relaxes the restriction):

> +A `super()` call must also be the first statement in the constructor of a derived class when the derived class contains initialized properties or has parameter properties.
> +Neither `Horse` or `Snake`, however, have those kind of properties right now.
> +So, as long as those two classes stay that way, they could have console logging or other statements before the `super()` call in their constructors without generating compile-time errors.

pdfernhout:
@sandersn  I've brought the pull request back to just moving the sentence based on [discussion for a now outdated diff](https://github.com/Microsoft/TypeScript-Handbook/pull/214#discussion_r61071435) with @mhegazy. Perhaps we can move forward with integrating this less-controversial change which is clearly needed and then discuss other incremental enhancements to the constructor explanation in another issue?

sandersn:
Sure, sorry about the wild goose chase. Thanks for your patience on this one.

pdfernhout:
@sandersn Thanks for merging the commit. I learned more about TypeScript and ES6 in the process of working on this, so the chase was good exercise. :-) 

Also, sometimes when things are challenging to document, that means part of a design could be improved. So, perhaps TypeScript itself will improve further from the issue opened on relaxing constructor rules, which in turn would make it easier for us to improve the constructor documentation.

PR #215, opened 2016-03-28
Fix wording in Classes.md on name private member

pdfernhout:
This change fixes a grammatical error introduced in this commit: https://github.com/Microsoft/TypeScript-Handbook/commit/76cb268c82ef6c4474c80c74930ba744060259c0

The fix was made working from the assumption that the referenced commit intended to further shorten the wording in the changed sentence (like changing "it is" to "it's") but did not fully complete that task.

Feel free to reject this pull request, however, and restore the original longer wording for the modified sentence ("it is not the same one as the one we declared") if that is preferred.


sandersn:
Thanks @pdfernhout !

PR #216, opened 2016-03-28
Data types and defaults added; options sorted alphabetically

SetTrend:
Solved issue https://github.com/Microsoft/TypeScript/issues/6086


pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--experimentalDecorators`                     | `false`<br/> <br/>`<boolean>`                  | Enables experimental support for ES7 decorators.
+`--forceConsistentCasingInFileNames`           | `false`<br/> <br/>`<boolean>`                  | Disallow inconsistently-cased references to the same file.
+`--help`<br/>`-h`                              |                                                | Print help message.
+`--inlineSourceMap`                            | `false`<br/> <br/>`<boolean>`                  | Emit a single file with source maps instead of having a separate file.
+`--inlineSources`                              | `false`<br/> <br/>`<boolean>`                  | Emit the source alongside the sourcemaps within a single file; requires `--inlineSourceMap` or `--sourceMap` to be set.
+`--isolatedModules`                            | `false`<br/> <br/>`<boolean>`                  | Unconditionally emit imports for unresolved files.
+`--jsx`                                        | `"Preserve"`<br/> <br/>`<JsxEmit>`             | Support JSX in '.tsx' files: 'React' or 'Preserve'. See [JSX](./JSX.md).
```
mhegazy: "JsxEmit" does not help much here. it should be `"preserve" | "react"`

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--listFiles`                                  | `false`<br/> <br/>`<boolean>`                  | Print names of files part of the compilation.
+`--locale`                                     | *(platform specific)*<br/> <br/>`<string>`     | The locale to use to show error messages, e.g. en-us.
+`--mapRoot`                                    | `null`<br/> <br/>`<string>`                    | Specifies the location where debugger should locate map files instead of generated locations. Use this flag if the .map files will be located at run-time in a different location than than the .js files. The location specified will be embedded in the sourceMap to direct the debugger where the map files where be located.
```
mhegazy: `null` is not a default. i would leave it empty.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--module`<br/>`-m`                            | `(target === "ES6" ? "ES6" : "CommonJS")`<br/> <br/>`<ModuleKind>` | Specify module code generation: 'commonjs', 'amd', 'system', 'umd', or 'es2015'. Only 'amd' and 'system' can be used in conjunction with `--outFile`. The 'es2015' value may not be used when targeting ES5 or lower.
```
mhegazy: same comment for `ModuleKind` as for `JsxEmit` earlier.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--moduleResolution`                           | `"Classic"`<br/> <br/>`<ModuleResolutionKind>` | Determine how modules get resolved. Either 'node' for Node.js/io.js style resolution, or 'classic' (default).
```
mhegazy: and ModuleRsolutionKind.

mhegazy: Add a link to ModuleResolution topic in https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Module%20Resolution.md

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--noImplicitAny`                              | `false`<br/> <br/>`<boolean>`                  | Raise error on expressions and declarations with an implied 'any' type.
+`--noImplicitReturns`                          | `false`<br/> <br/>`<boolean>`                  | Report error when not all code paths in function return a value.
+`--noImplicitUseStrict`                        | `false`<br/> <br/>`<boolean>`                  | Do not emit `"use strict"` directives in module output.
+`--noLib`                                      | `false`<br/> <br/>`<boolean>`                  | Do not include the default library file (lib.d.ts).
+`--noResolve`                                  | `false`<br/> <br/>`<boolean>`                  | Do not add triple-slash references or module import targets to the list of compiled files.
+~~`--out`~~                                    |                                                | DEPRECATED. Use `--outFile` instead.
+`--outDir`                                     | `null`<br/> <br/>`<string>`                    | Redirect output structure to the directory.
```
mhegazy: do not use null

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--noFallthroughCasesInSwitch`                 | `false`<br/> <br/>`<boolean>`                  | Report errors for fallthrough cases in switch statement.
+`--noImplicitAny`                              | `false`<br/> <br/>`<boolean>`                  | Raise error on expressions and declarations with an implied 'any' type.
+`--noImplicitReturns`                          | `false`<br/> <br/>`<boolean>`                  | Report error when not all code paths in function return a value.
+`--noImplicitUseStrict`                        | `false`<br/> <br/>`<boolean>`                  | Do not emit `"use strict"` directives in module output.
+`--noLib`                                      | `false`<br/> <br/>`<boolean>`                  | Do not include the default library file (lib.d.ts).
+`--noResolve`                                  | `false`<br/> <br/>`<boolean>`                  | Do not add triple-slash references or module import targets to the list of compiled files.
+~~`--out`~~                                    |                                                | DEPRECATED. Use `--outFile` instead.
```
mhegazy: add the type for completenss

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--outDir`                                     | `null`<br/> <br/>`<string>`                    | Redirect output structure to the directory.
+`--outFile`                                    | `null`<br/> <br/>`<string>`                    | Concatenate and emit output to single file. The order of concatenation is determined by the list of files passed to the compiler on the command line along with triple-slash references and imports. See output file order documentation for more details.
```
mhegazy: no null.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--preserveConstEnums`                         | `false`<br/> <br/>`<boolean>`                  | Do not erase const enum declarations in generated code. See [const enums documentation](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#94-constant-enum-declarations) for more details.
+`--pretty`<sup>[1]</sup>                       | `false`<br/> <br/>`<boolean>`                  | Stylize errors and messages using color and context.
+`--project`<br/>`-p`                           | `"."`<br/> <br/>`<string>`                     | Compile the project in the given directory. The directory needs to contain a `tsconfig.json` file to direct compilation. See [tsconfig.json](./tsconfig.json.md) documentation for more details.
```
mhegazy: "." is not the default. there is no default if not specified.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--reactNamespace`                             | `"react"`<br/> <br/>`<string>`                 | Specifies the object invoked for `createElement` and `__spread` when targeting 'react' JSX emit.
```
mhegazy: there is no default here either.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--removeComments`                             | `false`<br/> <br/>`<boolean>`                  | Remove all comments except copy-right header comments beginning with `/*!`
+`--rootDir`                                    | `"."`<br/> <br/>`<string>`                     | Specifies the root directory of input files. Only use to control the output directory structure with `--outDir`.
```
mhegazy: the default is not correct. if rootDir is not specified, a common root directory is computed from the list of input files.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--skipDefaultLibCheck`                        | `false`<br/> <br/>`<boolean>`                  | Don't check a user-defined default lib file's valitidy.
+`--sourceMap`                                  | `false`<br/> <br/>`<boolean>`                  | Generates corresponding '.map' file.
+`--sourceRoot`                                 | `null`<br/> <br/>`<string>`                    | Specifies the location where debugger should locate TypeScript files instead of source locations. Use this flag if the sources will be located at run-time in a different location than that at design-time. The location specified will be embedded in the sourceMap to direct the debugger where the source files where be located.
```
mhegazy: do not use null.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--strictNullChecks`                           | `false`<br/> <br/>`<boolean>`                  | In strict null checking mode, the `null` and `undefined` values are not in the domain of every type and are only assignable to themselves and `any` (the one exception being that `undefined` is also assignable to `void`).
+`--stripInternal`<sup>[1]</sup>                | `false`<br/> <br/>`<boolean>`                  | Do not emit declarations for code that has an `/** @internal */` JSDoc annotation.
+`--suppressExcessPropertyErrors`<sup>[1]</sup> | `false`<br/> <br/>`<boolean>`                  | Suppress excess property checks for object literals.
+`--suppressImplicitAnyIndexErrors`             | `false`<br/> <br/>`<boolean>`                  | Suppress `--noImplicitAny` errors for indexing objects lacking index signatures. See [issue #1232](https://github.com/Microsoft/TypeScript/issues/1232#issuecomment-64510362) for more details.
+`--target`<br/>`-t`                            | `"ES5"`<br/> <br/>`<ScriptTarget>`             | Specify ECMAScript target version: 'ES3' (default), 'ES5', or 'ES6'<sup>[1]</sup>
```
mhegazy: Same comment for ScriptTarget as for JSXEmit earlier.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--sourceRoot`                                 | `null`<br/> <br/>`<string>`                    | Specifies the location where debugger should locate TypeScript files instead of source locations. Use this flag if the sources will be located at run-time in a different location than that at design-time. The location specified will be embedded in the sourceMap to direct the debugger where the source files where be located.
+`--strictNullChecks`                           | `false`<br/> <br/>`<boolean>`                  | In strict null checking mode, the `null` and `undefined` values are not in the domain of every type and are only assignable to themselves and `any` (the one exception being that `undefined` is also assignable to `void`).
+`--stripInternal`<sup>[1]</sup>                | `false`<br/> <br/>`<boolean>`                  | Do not emit declarations for code that has an `/** @internal */` JSDoc annotation.
+`--suppressExcessPropertyErrors`<sup>[1]</sup> | `false`<br/> <br/>`<boolean>`                  | Suppress excess property checks for object literals.
```
mhegazy: This should not be marked as experimental. remove the `[1]`

pages/Compiler Options.md
mhegazy: should not be marked as experimental. remove the `[1]`

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--mapRoot`                                    | `null`<br/> <br/>`<string>`                    | Specifies the location where debugger should locate map files instead of generated locations. Use this flag if the .map files will be located at run-time in a different location than than the .js files. The location specified will be embedded in the sourceMap to direct the debugger where the map files where be located.
+`--module`<br/>`-m`                            | `(target === "ES6" ? "ES6" : "CommonJS")`<br/> <br/>`<ModuleKind>` | Specify module code generation: 'commonjs', 'amd', 'system', 'umd', or 'es2015'. Only 'amd' and 'system' can be used in conjunction with `--outFile`. The 'es2015' value may not be used when targeting ES5 or lower.
+`--moduleResolution`                           | `"Classic"`<br/> <br/>`<ModuleResolutionKind>` | Determine how modules get resolved. Either 'node' for Node.js/io.js style resolution, or 'classic' (default).
+`--newLine`                                    | *(platform specific)*<br/> <br/>`<string>`     | Use the specified end of line sequence to be used when emitting files: 'CRLF' (dos) or 'LF' (unix)."
```
mhegazy: it is only `"CRLF" | "LF"`

mhegazy:
I wounder how it looks if you split the type and default in different columns?

SetTrend:
Good call.

I actually had this data split into two columns before (see history). ... The problem I was facing was that GitHub tables seem to have a fixed width. The right-most column containing the description then became so small that only three words of a description fitted in a row. That's why I chose to put the two values into a single column and stack them vertically.

Your suggestions are very welcome. I'm not determined in the layout at all. If you're confirming to use a certain layout, I'll gladly change it accordingly :grinning:

SetTrend:
I intentionally put the `null` default values into some of the cells to distinguish that information from the impression that the default would not be known and still to fill-in to the table.

pages/Compiler Options.md
SetTrend: Which line? I can't tell.

pages/Compiler Options.md
SetTrend: It's actualy:
```javascript
enum NewLineKind {
        CarriageReturnLineFeed = 0,
        LineFeed = 1,
    }
```

SetTrend:
After giving it another thought on stating the `enum` data types I believe you are right.

Would you want me to change the `enum` data types to `string` and make sure all enumeration values are given in the corresponding description column?

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
-`--noImplicitReturns`                   |           |  Report error when not all code paths in function return a value.
-`--noFallthroughCasesInSwitch`          |           |  Report errors for fallthrough cases in switch statement.
-`--allowUnreachableCode`                |           |  Do not report errors on unreachable code.
-`--forceConsistentCasingInFileNames`    |           |  Disallow inconsistently-cased references to the same file.
-`--allowSyntheticDefaultImports`        |           |  Allow default imports from modules with no default export. This does not affect code emit, just typechecking.
-`--allowJs`                             |           |  Allow JavaScript files to be compiled.
-`--noImplicitUseStrict`                 |           |  Do not emit `"use strict"` directives in module output.
+Option (Shorthand)                             | Default / Type                                               | Description
+-----------------------------------------------|:------------------------------------------------------------:|----------------------------------------------------------------------
```
DanielRosenwasser: Don't make the middle column centered.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
-`--allowUnusedLabels`                   |           |  Do not report errors on unused labels.
-`--noImplicitReturns`                   |           |  Report error when not all code paths in function return a value.
-`--noFallthroughCasesInSwitch`          |           |  Report errors for fallthrough cases in switch statement.
-`--allowUnreachableCode`                |           |  Do not report errors on unreachable code.
-`--forceConsistentCasingInFileNames`    |           |  Disallow inconsistently-cased references to the same file.
-`--allowSyntheticDefaultImports`        |           |  Allow default imports from modules with no default export. This does not affect code emit, just typechecking.
-`--allowJs`                             |           |  Allow JavaScript files to be compiled.
-`--noImplicitUseStrict`                 |           |  Do not emit `"use strict"` directives in module output.
+Option (Shorthand)                             | Default / Type                                               | Description
```
DanielRosenwasser: Remove "(Shorthand)"

DanielRosenwasser:
:+1: split into two columns.

mhegazy:
> Agree on splitting into two columns.

looking at it again, it could get crowded. i wounder if instead we do something like:

Option | Type | Description
----------|-------|-------------------
`--allowJs` | boolean | Allow JavaScript files to be compiled. </br> Default: false
`--jsx`        |`"Preserve" | "React"`  | Support JSX in '.tsx' files: 'React' or 'Preserve'. See [JSX](./JSX.md). </br> Default: "Preseve"

DanielRosenwasser:
That works but it gets repetitive with the default. Here's what it looks like with two columns.

Option | Type | Default | Description
----------|-------|----------|---------
`--allowJs` | boolean | false | Allow JavaScript files to be compiled.
`--jsx`        | "Preserve" \| "React" | "Preserve" | Support JSX in '.tsx' files: 'React' or 'Preserve'. See [JSX](./JSX.md).

I'm okay with either.

mhegazy:
The problem is that with two columns you do not get much space, and the text wraps. i am guessing this is why @SetTrend chose to keep them in the same column with a line break.

SetTrend:
Yepp. But I also kept the extra shortcut column, too. I removed the Shortcut and the Default column to keep the Description legible.

But I'll do it anything you like. Just decide and I'll do it your way. I just wouldn't want to go back and forth :wink:

mhegazy:
my vote would be for default with discription

SetTrend:
@mhegazy & @DanielRosenwasser 
Want to have a look at the updated version?

I'm not sure which options isn't experimental anymore now. Would you mind clarifying?

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
-`--allowUnreachableCode`                |           |  Do not report errors on unreachable code.
-`--forceConsistentCasingInFileNames`    |           |  Disallow inconsistently-cased references to the same file.
-`--allowSyntheticDefaultImports`        |           |  Allow default imports from modules with no default export. This does not affect code emit, just typechecking.
-`--allowJs`                             |           |  Allow JavaScript files to be compiled.
-`--noImplicitUseStrict`                 |           |  Do not emit `"use strict"` directives in module output.
+Option                                         | Type      | Default                        | Description
+-----------------------------------------------|-----------|--------------------------------|----------------------------------------------------------------------
+`--allowJs`                                    | `boolean` | `true`                         | Allow JavaScript files to be compiled.
+`--allowNonTsExtensions`                       | `boolean` | `true`                         | Allow files with file extensions other than ".ts" to be compiled.
```
mhegazy: this is not a compiler option

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
-`--noFallthroughCasesInSwitch`          |           |  Report errors for fallthrough cases in switch statement.
-`--allowUnreachableCode`                |           |  Do not report errors on unreachable code.
-`--forceConsistentCasingInFileNames`    |           |  Disallow inconsistently-cased references to the same file.
-`--allowSyntheticDefaultImports`        |           |  Allow default imports from modules with no default export. This does not affect code emit, just typechecking.
-`--allowJs`                             |           |  Allow JavaScript files to be compiled.
-`--noImplicitUseStrict`                 |           |  Do not emit `"use strict"` directives in module output.
+Option                                         | Type      | Default                        | Description
+-----------------------------------------------|-----------|--------------------------------|----------------------------------------------------------------------
+`--allowJs`                                    | `boolean` | `true`                         | Allow JavaScript files to be compiled.
```
mhegazy: default is not true, default is false.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--allowNonTsExtensions`                       | `boolean` | `true`                         | Allow files with file extensions other than ".ts" to be compiled.
+`--allowSyntheticDefaultImports`               | `boolean` | `(module === "system")`        | Allow default imports from modules with no default export. This does not affect code emit, just typechecking.
+`--allowUnreachableCode`                       | `boolean` | `true`                         | Do not report errors on unreachable code.
+`--allowUnusedLabels`                          | `boolean` | `true`                         | Do not report errors on unused labels.
```
mhegazy: default is not true, default is false.

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--charset`                                    | `string`  | `"utf8"`                       | The character set of the input files.
+`--declaration`<br/>`-d`                       | `boolean` | `false`                        | Generates corresponding '.d.ts' file.
+`--diagnostics`                                | `boolean` | `false`                        | Show diagnostic information.
+`--emitBOM`                                    | `boolean` | `false`                        | Emit a UTF-8 Byte Order Mark (BOM) in the beginning of output files.
+`--emitDecoratorMetadata`<sup>[1]</sup>        | `boolean` | `false`                        | Emit design-type metadata for decorated declarations in source. See [issue #2577](https://github.com/Microsoft/TypeScript/issues/2577) for details.
+`--experimentalDecorators`                     | `boolean` | `false`                        | Enables experimental support for ES7 decorators.
```
mhegazy: this is experimental

pages/Compiler Options.md
```diff
@@ -1,56 +1,60 @@
...
+`--stripInternal`<sup>[1]</sup>                | `boolean` | `false`                        | Do not emit declarations for code that has an `/** @internal */` JSDoc annotation.
+`--suppressExcessPropertyErrors`               | `boolean` | `false`                        | Suppress excess property checks for object literals.
+`--suppressImplicitAnyIndexErrors`             | `boolean` | `false`                        | Suppress `--noImplicitAny` errors for indexing objects lacking index signatures. See [issue #1232](https://github.com/Microsoft/TypeScript/issues/1232#issuecomment-64510362) for more details.
+`--target`<br/>`-t`                            | `string`  | `"ES5"`                        | Specify ECMAScript target version: `'es3'` (default), `'es5'`, or `'es6'`.
+`--traceModuleResolution`                      | `boolean` | `false`                        | Report module resolution log messages.
```
mhegazy: traceResolution now.

mhegazy:
consider removing the defaults if they are "false" and/or "null".

SetTrend:
Thanks for sharing your comments with me.

Admittedly, I don't understand a word from what the green/red areas contain:

github

These code snippets are just gibberish to me.

Would you mind stating in prose which option you are referring to in your commments?

pages/Compiler Options.md
mhegazy: the default for allowUnusedLabels is false.

pages/Compiler Options.md
mhegazy: default for allowJs is false.

pages/Compiler Options.md
mhegazy: allowNonTsExtensions is not a compiler option

pages/Compiler Options.md
mhegazy: experimentalDecorators is experimental

pages/Compiler Options.md
mhegazy: rename `--traceModuleResolution` to `--traceResolution`

mhegazy:
added more details to comments.

pages/Compiler Options.md
SetTrend: Given the following line of code:

https://github.com/Microsoft/TypeScript/blob/master/src/server/editorServices.ts#l1359

Isn't it true, that `allowJs` is **false** if a `tsconfig.json` file exists omitting `allowJs`,
yet `allowJs` is **true** if _no_ `tsconfig.json` file exists?

So, doesn't it actually have two different default values, depending on whether a configuration file exists?

Would you mind checking and clarifying? I suggest to add a corresponding default and description then.

SetTrend:
> consider removing the defaults if they are "false" and/or "null".

I feel it's leaving the reader confused when a default value is lacking. I would ask myself: "What's the actual default now?" - "Couldn't the author decide on what the default is?" - "Is it random?"

I believe that `false` is a valid default value, as well as `true` is. Same for `null`. In fact, they are the actual default values. (Well, `null` should actually be `undefined`, though.)

mhegazy:
thanks!

PR #217, opened 2016-03-28
Rename 'definition file' to 'declaration file'

sandersn:
Also update the content of the file to be a little more consistent. The rest of documentation still uses the terms interchangeably.


DanielRosenwasser:
We'll need to include a redirect on the handbook as a heads up.

DanielRosenwasser:
:+1:

sandersn:
OK, I will create a redirecting `writing-definition-files.html` inside _docs.

PR #222, opened 2016-04-04
Fix typo in type-compatibility

sccolbert:
This adds a missing predicate to the introductory sentence of "Comparing two functions". Without it, the sentence is grammatically incorrect.


mhegazy:
Thanks

PR #224, opened 2016-04-05
Update --project flag tsconfig.json.md to have two dashes as per spec

xiphiaz:
Updates docs to say `--project` instead of `-project` (which is actually valid, but inconsistent with cli help and documentation elsewhere)


mhegazy:
@xiphiaz can you update the PR?

xiphiaz:
@mhegazy yup done

mhegazy:
thanks!

PR #229, opened 2016-04-07
Fix compile errors in Advanced Types.md examples.

gburdeti:
Fixes #

- Change padder type to Padder so there are no type errors.
- Use a type assertion for result, first and second to avoid compile errors.
- Add missing ConsoleLogger class implementing Logger for the mixin example.


DanielRosenwasser:
Thanks!

PR #231, opened 2016-04-07
Gulp quick start

sandersn:
Covers basic gulp, then adds browserify, then shows how to add watchify, uglify or babelify to the basic gulp+browserify pipeline.


pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+To start, we're going to structure our project in the following way:
+
+```text
+proj/
+    +- src/
+    +- dist/
+```
+
+TypeScript files will start out in your `src` folder, run through the TypeScript compiler and end up in `dist`.
```
DanielRosenwasser: So the thing about `dist` is that it's typically a convention for bundled packages. Otherwise I think things are usually placed in `lib`. We can talk about it offline.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+var gulp = require('gulp');
+var ts = require('gulp-typescript');
+var paths = {
+    scripts: ['src/**/*.ts'],
+};
+
+
+gulp.task('default', function () {
+    return gulp.src(paths.scripts)
+        .pipe(ts({noImplicitAny: true}))
```
DanielRosenwasser: I would create a variable named `tsOptions` and feed that into this.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        .pipe(gulp.dest('dist'));
```
DanielRosenwasser: Double quotes to be consistent with our other guides

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+gulp
+cd dist
+node main.js
+```
+
+The program should print "Hello from TypeScript!".
+
+# Browserify
+
+Now let's turn our code into a web app written as a collection of modules.
```
DanielRosenwasser: Consider

> Now let's build our code out and add a few modules to the mix.

DanielRosenwasser: I would consider showing using multiple modules and targeting `lib` first, then moving that towards using Browserify.

DanielRosenwasser: Never mind, you do this; however, you might still want to target `lib` instead of `dist`.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+Before we get to Browserify, let's make sure that the modules work by running `gulp` and then testing in Node:
+
+```shell
+gulp
+cd dist
+node main.js
+```
+
+Notice that even though we used ES2015 module syntax, TypeScript emitted CommonJS modules that Node uses.
+You can change this by adding a `module` member to the options object that you pass to gulp-typescript.
```
DanielRosenwasser: > by setting the `module` setting in the options that we passed earlier to gulp-typescript.

pages/quick-start/gulp.md
DanielRosenwasser: Actually, I'd still move this section out because it's independent of Browserify, and you talk about using AMD.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+For example `module: "amd` emits AMD module syntax instead.
+
+## Change to a web app
+
+Now let's move this from Node to the browser.
+We'll use browserify, which bundles all our modules into one JavaScript file.
```
DanielRosenwasser: Tell the user that to do this, they need to target CommonJS.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+In addition, it lets us use the CommonJS module system used by Node, which is the default TypeScript emit.
+That means we don't have to change any options to tell TypeScript which module system to target.
```
DanielRosenwasser: > If earlier on you switched your module type away from CommonJS, you should change it back.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+First, install browserify, [tsify](https://www.npmjs.com/package/tsify) and vinyl-source-stream.
+tsify is a browserify plugin that, like gulp-typescript, gives access to the TypeScript compiler.
+vinyl-source-stream lets us adapt the file output of browserify back into a gulp stream.
```
DanielRosenwasser: Browserify

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+## Change to a web app
+
+Now let's move this from Node to the browser.
+We'll use browserify, which bundles all our modules into one JavaScript file.
+In addition, it lets us use the CommonJS module system used by Node, which is the default TypeScript emit.
+That means we don't have to change any options to tell TypeScript which module system to target.
+
+First, install browserify, [tsify](https://www.npmjs.com/package/tsify) and vinyl-source-stream.
+tsify is a browserify plugin that, like gulp-typescript, gives access to the TypeScript compiler.
```
DanielRosenwasser: Browserify

pages/quick-start/gulp.md
DanielRosenwasser: back into a format that gulp understands called [Vinyl](https://github.com/gulpjs/vinyl).

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        <script src="bundle.js"></script>
+    </body>
+</html>
+```
+
+Now change `main.ts` to update the page:
+
+```ts
+import { sayHello, showHello } from "./greet";
+console.log(sayHello("TypeScript"));
```
DanielRosenwasser: Add a newline above `console.log`

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+showHello("greeting");
```
DanielRosenwasser: I'm not super big on the idea of `showHello` modifying the page and `sayHello` giving back a `string`. It's kind of confusing.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+Calling `showHello` makes `main.ts` change the paragraph's text in addition to writing to the console.
```
DanielRosenwasser: `showHello` only modifies the page's text, it doesn't write to the console

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        packageCache: {}
+    })
+    .plugin(tsify, {noImplicitAny: true})
+    .bundle()
+    .pipe(source('bundle.js'))
+    .pipe(gulp.dest('dist'));
+});
+```
+
+This adds the `copyHtml` task and changes the `default` task significantly.
```
DanielRosenwasser: Is there a convention here? My understanding is that these are usually written in kebab case, but I might be wrong

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+`copyHtml` is a simple task that needs to run before `default`.
```
DanielRosenwasser: I think you could change the last two lines to

> This adds the `copyHtml` task and adds it as a dependency of the `default`.
That means that any time `default` is run, `copyHtml` has to be run first.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+`default` now calls browserify first instead of using the usual gulp methods.
```
DanielRosenwasser: We've also changed `default` so that it now calls Browserify instead of gulp-typescript.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+After calling `bundle` we use `source` (our alias for vinyl-source-stream) to name the bundle `bundle.js`.
```
DanielRosenwasser: Weirdly this doesn't really mention the `tsify` plugin.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+Test the page by running gulp and then opening `dist/index.html` in a browser.
+You should see "Hello from TypeScript" on the page.
+
+Notice that we specify `debug: true` to browserify.
```
DanielRosenwasser: specified

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+This causes tsify to emit inline source maps.
```
DanielRosenwasser: This tells tsify to emit source maps inline in its `js` file output, which allows you to debug your bundled files as if you were debugging your own TypeScript code.

DanielRosenwasser: That might be a bit wordy, but it's not obvious what source maps are.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: I didn't intend to have people change to AMD. I was just giving that as an example. Since it's confusing, should I remove the mention entirely or instead try to make it clear that it's just a hypothetical example?

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: maybe one that logs and one that modifies the page?

pages/quick-start/gulp.md
sandersn: fixed

pages/quick-start/gulp.md
sandersn: let me find out

sandersn: Looking at other tutorials, there's a lot of variation between kebab-case, mediaevalcase and camelCase. kebab-case seems slightly more common (and better) so I'll use that.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: reworded

pages/quick-start/gulp.md
sandersn: Added some discussion.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: Good idea. Done.

sandersn:
@DanielRosenwasser can you take a look at the new commit when you have a chance?

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,397 @@
...
+
+The program should print "Hello from TypeScript!".
+
+# Browserify
+
+Now let's build our code out and add modules to the mix.
+This is the structure you're more likely to use for a real web app.
+We'll use Browserify to pack our code into a single file for distribution.
+
+## Add modules to the code
```
DanielRosenwasser: I still think you should add modules to your app even before mentioning Browserify.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,397 @@
...
+
+Create a file called `src/greet.ts`:
+
+```ts
+export function showHello(divName: string, name: string) {
+    const elt = document.getElementById(divName)
+    elt.innerText = `Hello from ${name}`;
+}
+export function sayHello(name: string) {
```
DanielRosenwasser: Newline above

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,397 @@
...
+Before we get to Browserify, let's make sure that the modules work by running `gulp` and then testing in Node:
+
+```shell
+gulp
+cd dist
+node main.js
+```
+
+Notice that even though we used ES2015 module syntax, TypeScript emitted CommonJS modules that Node uses.
+We'll stick with CommonJS for this tutorial, but you could set `module` in the options object to change this.
```
DanielRosenwasser: The way Browserify works is that you need to target CommonJS, which is why I still think you should split this out of the Browserify section

pages/quick-start/gulp.md
sandersn: OK, moved the module conversion to the previous section.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done, published in the new commit

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+    .bundle()
+    .pipe(source('bundle.js'))
+    .pipe(gulp.dest("dist"));
+});
+```
+
+This adds the `copy-html` task and adds it as a dependency of `default`.
+That means any time `default` is run, `copy-html` has to run first.
+We've also changed `default` to call Browserify first instead of gulp-typescript.
+Then we use the tsify plugin for Browserify instead of gulp-typescript.
```
DanielRosenwasser: > We've also changed `default` to call Browserify with the tsify plugin instead of gulp-typescript.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+Thankfully they both allow us to pass an options object to the TypeScript compiler.
```
DanielRosenwasser: Perhaps

> Conveniently, they both allow us to pass the same options object in this case.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+After calling `bundle` we use `source` (our alias for vinyl-source-stream) to name the bundle `bundle.js`.
```
DanielRosenwasser: our output bundle `bundle.js`

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+Test the page by running gulp and then opening `dist/index.html` in a browser.
+You should see "Hello from TypeScript" on the page.
+
+Notice that we specified `debug: true` to browserify.
```
DanielRosenwasser: Browserify

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+This causes tsify to emit source maps inside the bundled JavaScript file.
+Source maps let you debug the original TypeScript in the browser instead of the bundled JavaScript.
```
DanielRosenwasser: your original TypeScript code

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+You can test that source maps are working by opening the debugger for your browser and putting a breakpoint inside `main.ts`.
+When you refresh the page the breakpoint should pause the page and let you debug into `greet.ts`.
```
DanielRosenwasser: Remove "into"

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+# Watchify, Babel and Uglify
+
+Now that we are bundling our code with browserify and tsify, we can add various features to our build with browserify plugins.
```
DanielRosenwasser: Browserify

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+* Watchify starts gulp and keeps it running, incrementally compiling whenever you save a file.
+  This lets you keep an edit-save-refresh cycle going in the browser.
+
+* Babel is a hugely flexible transpiler that converts ES2015 and beyond into ES5 and ES3.
```
DanielRosenwasser: Do we have to call it a transpiler? :smiley:

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+  This lets you add transformations that TypeScript doesn't support.
```
DanielRosenwasser: Perhaps

> This lets enables more extensive transformations that TypeScript might not support.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+* Uglify compacts your code so that it takes less time to download onto the client.
```
DanielRosenwasser: > so that it holds a smaller footprint and takes less time to download.

In theory, you could distribute an Uglified package on npm

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+var source = require('vinyl-source-stream');
+var watchify = require("watchify");
+var tsify = require("tsify");
+var paths = {
+    pages: ['*.html']
+};
+var tsOptions = { noImplicitAny: true };
+
+const b = watchify(browserify({
+        basedir: '.',
```
DanielRosenwasser: Why not 4-space indentation? Seems odd even within a nested call.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        debug: true,
+        entries: ['src/main.ts'],
+        cache: {},
+        packageCache: {}
+})
+        .plugin(tsify, tsOptions));
```
DanielRosenwasser: This looks even odder

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+gulp.task("copy-html", function () {
```
DanielRosenwasser: Newline above

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+var browserify = require("browserify");
+var source = require('vinyl-source-stream');
+var watchify = require("watchify");
+var tsify = require("tsify");
+var paths = {
+    pages: ['*.html']
+};
+var tsOptions = { noImplicitAny: true };
+
+const b = watchify(browserify({
```
DanielRosenwasser: `const` might not be supported

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        entries: ['src/main.ts'],
+        cache: {},
+        packageCache: {}
+})
+        .plugin(tsify, tsOptions));
+gulp.task("copy-html", function () {
+    return gulp.src(paths.pages)
+        .pipe(gulp.dest("dist"));
+});
+function bundle() {
```
DanielRosenwasser: newline above

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+    return b
+        .bundle()
+        .pipe(source('bundle.js'))
+        .pipe(gulp.dest("dist"));
+}
+gulp.task("default", ["copy-html"], bundle);
```
DanielRosenwasser: newline above

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+b.on("update", bundle);
+```
+
+There are basically two changes here, but they require you to refactor your code a bit.
+
+1. Wrap the `browserify` plus `tsify` call in a `watchify` call.
```
DanielRosenwasser: What about

> We wrapped our browserify instance in a call to `watchify`, and then held onto its result.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+   Then save the resulting object in a variable named `b`.
```
DanielRosenwasser: Let's use something better-named than `b`.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+2. Call `b.on("update", bundle);` to run browserify's bundling every time one of your TypeScript files changes.
```
DanielRosenwasser: We cal

DanielRosenwasser: so that Browserify will run the `bundle` function every time one of your TypeScript files changes.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+Together (1) and (2) mean that you have to separate the browserify construction out of the `default` task.
```
DanielRosenwasser: we have to move our call to `browserify` out of the `default` task

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+And you have to give the function for `default` a name since you need it both for the `gulp.task` and for `watchify.on`.
```
DanielRosenwasser: We then factored out the default task into a standalone function so that it could be used as both the `default` task handler and as the callback for a change in our input files.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        .pipe(source('bundle.js'))
+        .pipe(buffer())
+        .pipe(sourcemaps.init({loadMaps: true}))
+        .pipe(uglify())
+        .pipe(sourcemaps.write('./'))
+        .pipe(gulp.dest("dist"));
+});
+```
+
+Notice that `uglify` itself just one call &mdash; the calls to `buffer` and `sourcemaps` exist to make sure sourcemaps keep working.
```
DanielRosenwasser: \[has or uses] just one call

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+These calls give us a separate sourcemap file instead of using inline sourcemaps like before.
+
+## Babel
+
+First install Babelify.
+Like Uglify, Babelify mangles code, so we'll need vinyl-buffer and gulp-sourcemaps.
```
DanielRosenwasser: This isn't quite the intention, but it's not untrue. :smile:

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        .pipe(source('bundle.js'))
+        .pipe(buffer())
+        .pipe(sourcemaps.init({loadMaps: true}))
+        .pipe(uglify())
+        .pipe(sourcemaps.write('./'))
+        .pipe(gulp.dest("dist"));
+});
+```
+
+Notice that TypeScript now targets ES6.
```
DanielRosenwasser: ES2015

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+var browserify = require("browserify");
+var source = require('vinyl-source-stream');
+var tsify = require("tsify");
+var uglify = require('gulp-uglify');
+var sourcemaps = require('gulp-sourcemaps');
+var buffer = require('vinyl-buffer');
+var paths = {
+    pages: ['*.html']
+};
+var tsOptions = { noImplicitAny: true };
```
DanielRosenwasser: I think you accidentally copied/pasted the above gulpfule - this doesn't seem to demonstrate Babelify

sandersn:
Also we discussed in person that adding "You might need `sudo`" after the first `npm install` would be nice for people on Unix.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: hmm, I was thinking "step into" when I wrote it, but just "debug" is actually clearer.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: I thought I did until I re-checked Babel's home page. They call themselves a compiler, so I'll go for that.

pages/quick-start/gulp.md
sandersn: Maybe. I think it's worthwhile to emphasize that Babel gives you basically any transformation you can imagine, but I think it's also a strong contrast to TypeScript's straightforward (1) type erasure  (2) stage 3+ proposals. I changed it to:

"This lets you add extensive and customized transformations that TypeScript does't support."

pages/quick-start/gulp.md
sandersn: Maybe, but this is a one-liner summary so I want to get the main usage here.

sandersn: removed "onto the client".

pages/quick-start/gulp.md
sandersn: left over from various edits. 4-space makes a lot more sense.

pages/quick-start/gulp.md
sandersn: Yeah, I was trying to ignore `js-mode`'s suggestion to line it up under the 'b' of `browserify`. Maybe I'll just join with the previous line to get 

```js
}).plugin(tsify, tsOptions));
```

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: good catch. Funny how fast I got used to using that keyword.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: `watchedBrowserify` ?

pages/quick-start/gulp.md
sandersn: nice

pages/quick-start/gulp.md
sandersn: I went with the even simpler "And we have to give the function for `default` a name since both Watchify and Gulp need to call it."

pages/quick-start/gulp.md
sandersn: I'm not completely sold, but ok.

pages/quick-start/gulp.md
sandersn: done -- also changed the tsify options to reference es2015 instead of es6.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,399 @@
...
+
+Now we can use `npm install` to install packages.
+First install TypeScript and gulp globally.
+(You might need to start `npm install` commands in this guide with `sudo` if you're on a Unix system.)
+
+```shell
+npm install -g typescript gulp-cli
+```
+
+Then install `gulp` and `gulp-typescript` in your project `devDependencies`.
```
DanielRosenwasser: project's

DanielRosenwasser: consider: dev dependencies

DanielRosenwasser:
I just realized that this doesn't include a `tsconfig.json` =/

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+hello("TypeScript");
+```
+
+### Create a `gulpfile.js`
+
+In the project root, create the file `gulpfile.js`:
+
+```js
+var gulp = require("gulp");
+var ts = require('gulp-typescript');
```
DanielRosenwasser: Mixed quotes here

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+        .pipe(gulp.dest("dist"));
+});
+```
+
+### Test the resulting app
+
+```shell
+gulp
+cd dist
+node main.js
```
DanielRosenwasser: Remove the `cd` and just make it `node dist/main.js`

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+
+console.log(sayHello("TypeScript"));
+```
+
+Make sure that the modules work by running `gulp` and then testing in Node:
+
+```shell
+gulp
+cd dist
+node main.js
```
DanielRosenwasser: Remove the `cd` and just make it `node dist/main.js`

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+Notice that even though we used ES2015 module syntax, TypeScript emitted CommonJS modules that Node uses.
+We'll stick with CommonJS for this tutorial, but you could set `module` in the options object to change this.
+
+# Browserify
+
+Now let's move this project from Node to the browser.
+To do this, we need to bundle all our modules into one JavaScript file.
```
DanielRosenwasser: "we need to" -> "we'd like to"

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,399 @@
...
+        <script src="bundle.js"></script>
+    </body>
+</html>
+```
+
+Now change `main.ts` to update the page:
+
+```ts
+import { sayHello } from "./greet";
+export function showHello(divName: string, name: string) {
```
DanielRosenwasser: Newline above

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,399 @@
...
+```html
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8" />
+        <title>Hello World!</title>
+    </head>
+    <body>
+        <p id="greeting">Loading ...</p>
+        <script src="bundle.js"></script>
```
DanielRosenwasser: This needs to be `./dist/bundle.js`

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,399 @@
...
+tsify is a Browserify plugin that, like gulp-typescript, gives access to the TypeScript compiler.
+vinyl-source-stream lets us adapt the file output of Browserify back into a format that gulp understands called [vinyl](https://github.com/gulpjs/vinyl).
+
+```shell
+npm install --save-dev browserify tsify vinyl-source-stream
+```
+
+### Create a page
+
+Create a file in the root of the project named `index.html`:
```
DanielRosenwasser: Since you're funneling `index.html` to `dist`, maybe this should go in `src` as well.

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,399 @@
...
+Calling `showHello` calls `sayHello` to change the paragraph's text.
+Now change your gulpfile to the following:
+
+```js
+var gulp = require("gulp");
+var browserify = require("browserify");
+var source = require('vinyl-source-stream');
+var tsify = require("tsify");
+var paths = {
+    pages: ['*.html']
```
DanielRosenwasser: If you move `index.html` to `src`, you should make this `src/*.html`

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,399 @@
...
+watchedBrowserify.on("update", bundle);
+```
+
+There are basically two changes here, but they require you to refactor your code a bit.
+
+1. We wrapped our `browserify` instance in a call to `watchify`, and then held on to the result.
+2. We called `watchedBrowserify.on("update", bundle);` so that Browserify will run the `bundle` function every time one of your TypeScript files changes.
+
+Together (1) and (2) mean that we have to move our call to `browserify` out of the `default` task.
+And we have to give the function for `default` a name since both Watchify and Gulp need to call it.
```
DanielRosenwasser: Wow, I tried this out and there is zero feedback when Watchify completes. Can you add gulp-util as a dev dependency on this step and add a `watchedBrowserify.on('log', gutil.log);` as per [this page](https://github.com/gulpjs/gulp/blob/f4870fd83a04e5e72afa2be4ea0d32b2c70eb9c9/docs/recipes/fast-browserify-builds-with-watchify.md)?

DanielRosenwasser: Also, can you tell the user to try it out and try changing their files?

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,399 @@
...
+    .pipe(buffer())
+    .pipe(sourcemaps.init({loadMaps: true}))
+    .pipe(uglify())
+    .pipe(sourcemaps.write('./'))
+    .pipe(gulp.dest("dist"));
+});
+```
+
+Notice that `uglify` itself has just one call &mdash; the calls to `buffer` and `sourcemaps` exist to make sure sourcemaps keep working.
+These calls give us a separate sourcemap file instead of using inline sourcemaps like before.
```
DanielRosenwasser: Can you tell the user to check out their `dist/bundle.js` file and see how it has now been smushed?

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: I thought `src` was for code things, especially ones that get transformed. The knockout and react  guides do it this way too, so let's have them all place from the root.

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
sandersn: done

pages/quick-start/gulp.md
DanielRosenwasser: Right, but those include `js` files from `dist` off the bat. You're moving `index.html` from `src` to `dist`, which in my mind makes it an input file at that point.

DanielRosenwasser: Basically, the reason I'm inclined to change this is because I actually tried opening the root's `index.html` and was confused for a bit as to why it wasn't working.

sandersn: Fair enough

pages/quick-start/gulp.md
```diff
@@ -284,15 +284,34 @@ function bundle() {
...
 Together (1) and (2) mean that we have to move our call to `browserify` out of the `default` task.
 And we have to give the function for `default` a name since both Watchify and Gulp need to call it.
+Adding logging with (3) is optional but very useful for debugging your setup.
+
+Now when you run Gulp, it should start and stay running.
+Try changing the code for `showHello` in `main.ts` and saving it.
+You should see output that looks like this:
+
+```shell
+gulp-proj$ gulp
```
DanielRosenwasser: This should be `proj`

pages/quick-start/gulp.md
```diff
@@ -0,0 +1,456 @@
...
+
+In the project root, `proj`, create the file `tsconfig.json`:
+
+```json
+{
+    "files": [
+        "src/main.ts"
+    ],
+    "compilerOptions": {
+        "noImplicitAny": true
```
DanielRosenwasser:
```json
,
"target": "es5"
```

DanielRosenwasser:
Apart from my last comment, :+1:!

pages/quick-start/gulp.md
sandersn: done

PR #238, opened 2016-04-10
Various fixes to code examples having compilation errors.

gburdeti:
Fixes
- Specify obj as ```any``` to avoid a compilation error - Symbols Section.
- Add missing curly brace - Iterators and Generators Section - Iterables Subsection.
- Use a type assertion to ```any``` to avoid a compilation error - Iterators and Generators Section - Iterables Subsection.
- Call the RegExp constructor to avoid a compilation error - Modules Section.


gburdeti:
It's also possible to use
```ts
export const numberRegexp = /^[0-9]+$/;
```
for creating the regular expression instead of using the constructor.

pages/Iterators and Generators.md
```diff
@@ -32,14 +32,15 @@ for (let i in list) {
...
+}
 ```
 
 Another distinction is that `for..in` operates on any object; it serves as a way to inspect properties on this object.
 `for..of` on the other hand, is mainly interested in values of iterable objects. Built-in objects like `Map` and `Set` implement `Symbol.iterator` property allowing access to stored values.
 
 ```ts
 let pets = new Set(["Cat", "Dog", "Hamster"]);
-pets["species"] = "mammals";
+(<any>pets)["species"] = "mammals";
```
DanielRosenwasser: This is only an error with `noImplicitAny`

DanielRosenwasser: @mhegazy was it intended that you wanted to use a `Set` here? I don't even see an index signature on it.

pages/Modules.md
```diff
@@ -35,7 +35,7 @@ export interface StringValidator {
 ##### ZipCodeValidator.ts
 
 ```ts
-export const numberRegexp = /^[0-9]+$/;
+export const numberRegexp = new RegExp(/^[0-9]+$/);
```
DanielRosenwasser: Why was this an error?

pages/Iterators and Generators.md
mhegazy: I needed something that is itrable, map, set or array would do.

gburdeti:
Sorry I was checking the handbook on the website (typescriplang.org) and it was "/^[0-9]+$/" with the quotes and I just pasted the corrected line in the .md file from the repo without checking. My mistake. :(

pages/Iterators and Generators.md
mhegazy: I agree with @DanielRosenwasser the any more noise. The point of this example is to show the difference between for..in and for..of in JavaScript, do not assume someone would use this verbatim in their project.

gburdeti:
I made another pull request only for the minor typos. I reverted the version of the other files. Sorry again.

DanielRosenwasser:
Don't even worry about it, it's not a big deal. :smiley:

Issue #239, opened 2016-04-10
Indexing with symbols not yet supported - Symbols section


gburdeti:
The following code doesn't compile.

```ts
const getClassNameSymbol = Symbol();

class C {
    [getClassNameSymbol](){
       return "C";
    }
}

let c = new C();
let className = c[getClassNameSymbol](); // "C"
```
It's the last line that creates an error.

From what I can understand it is explained in issue [#1863](https://github.com/Microsoft/TypeScript/issues/1863) of the TypeScript repo.

Based on @danquirk's  and  **JsonFreeman**'s answers to **wereHamster** this feature is not yet supported.

gburdeti:
Nevermind, this code compiles correctly if *noImplicitAny* is set to false.

gburdeti:
Reopened based on comment from @sandersn in pull request #240 that examples need to be clean for "noImplicitAny": true.

PR #240, opened 2016-04-10
Add missing curly brace. Typo fix for a module import.

gburdeti:
Fixes #
- Change the variable name to *myValidator* to avoid a duplicate conflict - Modules Section.
- Typo fix for a module import using ```import =```.


pages/Symbols.md
```diff
@@ -24,7 +24,7 @@ Just like strings, symbols can be used as keys for object properties.
 ```ts
 let sym = Symbol();
 
-let obj = {};
+let obj: any = {};
```
mhegazy: I would change it to:

```ts
let obj = {
   [sym]: "value"
};

console.log(obj[sym]); // "value"
```

DanielRosenwasser: I had the same idea actually, good call.

DanielRosenwasser: Only thing is that we never explained computer properties :smile:

gburdeti:
@mhegazy made me realize that if *noImplicitAny* is set to false then the code works fine. I've set it in my tsconfig file. This kind of mistake won't happen again. Sorry.

sandersn:
Actually, we try to have the handbook examples clean for `"noImplicitAny": true`. Can you re-add the `any` assertions in `Advanced Types.md`?

sandersn:
After you add the casts back in, :+1:

mhegazy:
:+1:

gburdeti:
Then, I will reopen issue #239, because the code doesn't work unless *noImplicitAny* is false.

PR #241, opened 2016-04-11
Update asp-net-core.md

freezestudio:
Fixes #

change 'src' to 'scripts'

因为与上文中提到的不一致
Because it is not consistent with the above mentioned


sandersn:
Can you revert the change on line 8? The link title is actually correct. After that's fixed, :+1:

RyanCavanaugh:
No response

PR #243, opened 2016-04-12
Fix padLeft example

robinpokorny:
```js
Array(4).join(".") === "..."
```


sandersn:
Thanks @robinpokorny !

PR #253, opened 2016-04-18
Move pngs to assets

sandersn:
1. Move all pngs from pages/quick-starts/*png to assets/*png
2. Change in-page references from ./*png to ../assets/*png

This assumes that the quick-starts will be placed in quick-start/*html on the web site, not pages/quick-start/*html. @DanielRosenwasser can you comment on what will need to happen for that? It's mostly on the web site repo, right?


mhegazy:
@DanielRosenwasser any foreseen issues with the site?

mhegazy:
:+1:  otherwise

DanielRosenwasser:
I'd like a better separation between these assets. It's not clear which images are used by which pages. Also, I think we changed our minds and decided that these should be placed within `quick-starts/assets/`

sandersn:
As we discussed in person, let's put them in `quick-starts/assets/asp`

DanielRosenwasser:
I made the paths explicitly relative. I'd also rename the folder `aspnet` instead of `asp` but I think we can change that later on if we need.

PR #254, opened 2016-04-19
Adding a few missing words

### LINKED ISSUE #245: missing important word in explanation for Variable Declarations > Block-scoped variable capturing
"When we first touched on the idea of variable capturing with var declaration, we briefly went into how variables act once captured. To give a better intuition of this, each time a scope is run, it creates an “environment” of variables. That environment and its **[i dunno what goes here]** can exist even after everything within its scope has finished executing."

I'd do a pull request but I don't know what was intended be plugged in there.

drmacm:
Fixes #245 (and another 2 small unreported issues)

### DIFF
--- a/pages/Interfaces.md
+++ b/pages/Interfaces.md
@@ -196,7 +196,7 @@ mySearch = function(source: string, subString: string) {
 }
 ```
 
-For function types to correctly type-check, the name of the parameters do not need to match.
+For function types to correctly type-check, the name of the parameters does not need to match.
 We could have, for example, written the above example like this:
 
 ```ts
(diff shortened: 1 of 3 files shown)


pages/Interfaces.md
```diff
@@ -196,7 +196,7 @@ mySearch = function(source: string, subString: string) {
 }
 ```
 
-For function types to correctly type-check, the name of the parameters do not need to match.
+For function types to correctly type-check, the name of the parameters does not need to match.
```
DanielRosenwasser: The correct fix is actually to make it "names of the parameters"

DanielRosenwasser:
Thanks @drmacm, I just left one comment to address. Otherwise looks good.

drmacm:
Sure, I will fix it properly and resubmit immediately.

DanielRosenwasser:
Awesome, thanks for the PR!

PR #256, opened 2016-04-19
Possibly wrong formulation in Modules.md

### LINKED ISSUE #255: Possibly wrong formulation in Modules.md
In the section "export = and import = require()" there is a sentence:
"TypeScript supports export = to module the traditional CommonJS and AMD workflow."

I believe it should be:
"TypeScript supports export = to model the traditional CommonJS and AMD workflow."

PR coming in a minute.

drmacm:
Fixes #255


mhegazy:
thanks!

PR #261, opened 2016-04-27
Update Compiler Options.md

freezestudio:
tsc version 1.8.10

error TS5023: Unknown compiler option 'noerrortruncation'.


mhegazy:
thanks!

PR #262, opened 2016-04-27
Refactor images

bowdenk7:
This PR goes with the changes to the website (PR [here](https://github.com/Microsoft/TypeScript-Docs/pull/175)). It is necessary to make the URLs match up.


pages/tutorials/ASP.NET Core.md
```diff
@@ -1,4 +1,4 @@
-# ASP.NET Core
+# Setup
```
DanielRosenwasser: It's kinda odd that these had no intros but no big deal

DanielRosenwasser:
I thought this was all going to be in `pages/tutorials/assets` but I guess this makes more sense. Can you rename it to `React and Webpack`?

mhegazy:
Would this make the images show up correctelly here? i.e. on https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/quick-start/asp-net-core.md

DanielRosenwasser:
No because the problem is that these are not getting linked to their raw versions, they're getting linked to pages on GitHub.

PR #273, opened 2016-05-08
Noting example is below text and not above.

GrizzlyMcBear:
Fixes #

### DIFF
--- a/pages/Variable Declarations.md
+++ b/pages/Variable Declarations.md
@@ -45,7 +45,7 @@ var g = f();
 g(); // returns 11;
 ```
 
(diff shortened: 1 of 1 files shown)


mhegazy:
The original text seems accurate. it refers to the example above. what makes you believe it should refer to the following example instead?

DanielRosenwasser:
Yeah, but now that I look at it, I wonder what the below example is actually trying to show. Maybe that should be improved.

GrizzlyMcBear:
@mhegazy I thought the same at first, but after reading the second example
I saw that it had no description and also that the description seemed more relevant to it
instead of to the first example.

RyanCavanaugh:

PR #275, opened 2016-05-11
intended to use config.collor

materliu:
Fixes #

intended to use `config.collor` as the comment.


mhegazy:
the comment was intended for the next line:

```ts
newSquare.color = config.collor;
```

mhegazy:
may be we should move the error message to the following line instead.

PR #276, opened 2016-05-11
Double quotes

### LINKED ISSUE #94: Use single/double quotes consistently
The Handbook currently uses both single and double quotes, sometimes even in the same example:

```
var name: string = "bob";
name = 'smith';
```

Settling on just one style would look better.

wanderer06:
Fixes #94 

There are still some inconsistencies with the quoting style for entities inside code comments. I'm aware these should be single-quoted, but because they vary quite a lot (between quoted or not), I did not go into any great detail to avoid affecting any intended style.

Also, inside *template strings* the quotes are mostly single, with double quotes in some files. I'm not sure if this is also intended.

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -78,9 +78,9 @@ For example, you may want to represent a value as a pair of a `string` and a `nu
 // Declare a tuple type
 let x: [string, number];
 // Initialize it
-x = ['hello', 10]; // OK
+x = ["hello", 10]; // OK
 // Initialize it incorrectly
-x = [10, 'hello']; // Error
+x = [10, "hello"]; // Error
 ```
 
 When accessing an element with a known index, the correct type is retrieved:
@@ -93,11 +93,11 @@ console.log(x[1].substr(1)); // Error, 'number' does not have 'substr'
 When accessing an element outside the set of known indices, a union type is used instead:
 
 ```ts
-x[3] = 'world'; // OK, string can be assigned to (string | number)
+x[3] = "world"; // OK, string can be assigned to 'string | number'
 
 console.log(x[5].toString()); // OK, 'string' and 'number' both have toString
 
-x[6] = true; // Error, boolean isn't (string | number)
+x[6] = true; // Error, boolean isn't 'string | number'
 ```
 
 Union types are an advanced topic that we'll cover in a later chapter.
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -93,11 +93,11 @@ console.log(x[1].substr(1)); // Error, 'number' does not have 'substr'
 When accessing an element outside the set of known indices, a union type is used instead:
 
 ```ts
-x[3] = "world"; // OK, string can be assigned to 'string | number'
+x[3] = "world"; // OK, 'string' can be assigned to 'string | number'
 
-console.log(x[5].toString()); // OK, 'string' and 'number' both have toString
+console.log(x[5].toString()); // OK, 'string' and 'number' both have 'toString'
 
-x[6] = true; // Error, boolean isn't 'string | number'
+x[6] = true; // Error, 'boolean' isn't 'string | number'
 ```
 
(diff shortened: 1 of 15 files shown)


pages/Basic Types.md
```diff
@@ -93,11 +93,11 @@ console.log(x[1].substr(1)); // Error, 'number' does not have 'substr'
 When accessing an element outside the set of known indices, a union type is used instead:
 
 ```ts
-x[3] = 'world'; // OK, string can be assigned to (string | number)
+x[3] = "world"; // OK, string can be assigned to 'string | number'
 
 console.log(x[5].toString()); // OK, 'string' and 'number' both have toString
 
-x[6] = true; // Error, boolean isn't (string | number)
+x[6] = true; // Error, boolean isn't 'string | number'
```
DanielRosenwasser: Could you add single quotes around `boolean` and `string` above

pages/Variable Declarations.md
```diff
@@ -81,7 +81,7 @@ function f(shouldInitialize: boolean) {
     return x;
 }
 
-f(true);  // returns '10'
+f(true);  // returns 10
```
DanielRosenwasser: This change should be reverted

DanielRosenwasser:
I guess I'd need to see the template strings example but this looks great so far. I just left 2 comments for feedback.

pages/Variable Declarations.md
wanderer06: I've made that change to be consistent with the rest of the page.
Further up ahead, there are two instances where numbers are returned that are not quoted:
```
g(); // returns 11;
// ...
f(); // returns 2
```

I'll change both of these to be quoted then.

pages/Basic Types.md
wanderer06: Yes, that makes sense.

wanderer06:
@DanielRosenwasser updated commented files. Also regarding string templates:

**Advanced Types**
```
throw new Error(`Expected string or number, got '${value}'.`);
```

**Modules**
```
console.log(`"${s}" ${validate(s) ? " matches" : " does not match"}`);
console.log(`"${ s }" - ${ validator.isAcceptable(s) ? "matches" : "does not match" }`);

// ...
throw new Error(`Unsupported input: '${char}'`);
console.log(`result of '${input}' is '${c.getResult()}'`);
```

This is what lead me to believe it's intended. It seems to use single quotes to represent entities whenever an exception is thrown (i.e compiler error) and double when printing friendly information to the console to escape the quotes. As you would in something like:
```
console.log(`Hi "George"`); // console.log("Hi \"George\"");
```

But then that's not consistent in **Namespaces**:
```
console.log(`'${ s }' ${ isMatch ? "matches" : "does not match" } '${ name }'.`);
console.log(`"${ s }" - ${ validators[name].isAcceptable(s) ? "matches" : "does not match" } ${ name }`);
```

I hope I'm not being too pedantic!

pages/Functions.md
```diff
@@ -89,7 +89,7 @@ In playing with the example, you may notice that the TypeScript compiler can fig
 // myAdd has the full function type
 let myAdd = function(x: number, y: number): number { return  x + y; };
 
-// The parameters `x` and `y` have the type number
+// The parameters 'x' and 'y' have the type number
```
DanielRosenwasser: This should be `'number'`

DanielRosenwasser:
You're not being too pedantic - I think you've correctly captured the general intent. I'd feel bad leaving this PR hanging, so if you want we can pull everything in as-is and iterate. Otherwise, we can address that inconsistency in this PR. Any preference?

wanderer06:
Let's pull this in then, and raise a PR for other inconsistencies. 
I think mainly what's outstanding is how entities, return types, string literals etc are quoted in comments (inside code blocks).

DanielRosenwasser:
Thanks for the PR! Feel free to send fixes for any other issues you catch :smiley:

PR #277, opened 2016-05-12
Fix errors in Advanced Types

xcatliu:
```ts
throw new Error(`Expected string or number, got '${value}'.`);
```

probably should be

```ts
throw new Error(`Expected string or number, got '${padding}'.`);
```


DanielRosenwasser:
Thank you!

PR #285, opened 2016-05-22
Added let keyword to name declaration

carlosmunozrodriguez:
Is there any particular reason why the let keyword was omitted in this declaration?

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -32,7 +32,7 @@ Just like JavaScript, TypeScript also uses double quotes (`"`) or single quotes 
 
 ```ts
 let color: string = "blue";
-name = 'red';
+let name = 'red';
 ```
 
 You can also use *template strings*, which can span multiple lines and have embedded expressions.


carlosmunozrodriguez:
Oh I just realized `name` is already defined in `lib.d.ts`. Still this isn't explained in the docs. Maybe a different variable name would be better.

wanderer06:
@carlosmunozrodriguez I think it should still have the declaration. Either that or have it changed to:
```
let color: string = "blue";
color = 'red';
```

sandersn:
I think it should be changed to `color = 'red';`
It looks to me that switching to `name` was a mistake.

carlosmunozrodriguez:
Unfortunately using the keyword `let` generates this error

    error TS2451: Cannot redeclare block-scoped variable 'name'.

I think it is probably a bug to have `name` declared in `lib.d.ts`. Node has no problem running a JavaScript file with `let name = 'red';`

Anyway I like @wanderer06 suggestion of using

    let color: string = "blue";
    color = 'red';

DanielRosenwasser:
I don't know how I missed that, but thanks for pointing this out. I fixed the issue with f36b33012d7d03829cb7185964ceabe3a29daf8f a https://github.com/Microsoft/TypeScript-Handbook/blob/f36b33012d7d03829cb7185964ceabe3a29daf8f/pages/Basic%20Types.md#string.

I just figured a commit would be faster, but please feel free to send more PRs in the future. :smiley:

PR #293, opened 2016-06-08
Added Angular 2 tutorial page.

DanielRosenwasser:
Angular.io will always do a better job than we'll be able to keep up with. We're just going to forward users there.


pages/tutorials/Angular 2.md
```diff
@@ -0,0 +1,6 @@
+Angular 2 is an upcoming framework built in TypeScript.
+Using TypeScript with Angular is fairly easy to set up for this reason, as TypeScript fits right in.
```
mhegazy: TypeScript with **Angular 2**

mhegazy:
:+1:

sandersn:
:+1:

PR #296, opened 2016-06-10
Fix typo: Change the default value  for the target option

### LINKED ISSUE #295: Bad default value for the target compiler options
Right now, in this page https://www.typescriptlang.org/docs/handbook/compiler-options.html, 'ES3' is marked as the default value for the target option. That's not true anymore.

```
Specify ECMAScript target version: 'es3' (default), 'es5', or 'es6'.
```

Gillespie59:
Fixes #295


mhegazy:
as noted in https://github.com/Microsoft/TypeScript-Handbook/issues/295, the default stays ES3. there is no intention to change this at the time being.

PR #314, opened 2016-06-23
Null and Undefined

### LINKED ISSUE #313: Add documentation for null/undefined/never and strictnullchecks
null/undefiend/never to basic types
section about strictNullChecks

DanielRosenwasser:
Partially fixes #313.

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -206,7 +206,7 @@ That means you can assign `null` and `undefined` to something like `number`.
 
 However, when using the `--strictNullChecks` flag, `null` and `undefined` are only assignable to `void` and their respective types.
 This helps avoid *many* common errors.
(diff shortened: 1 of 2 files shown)


pages/Basic Types.md
```diff
@@ -190,6 +190,27 @@ Declaring variables of type `void` is not useful because you can only assign `un
...
+let u: undefined = undefined;
+let n: null = null;
+```
+
+By default `null` and `undefined` are subtypes of all other types.
+That means you can assign `null` and `undefined` to something like `number`.
+
+However, when using the `--strictNullChecks` flag, `null` and `undefined` are only assignable to `void` and their respective types.
+This helps avoid *many* common errors.
+In cases where you wanted to pass in either a `string` or `null` or `undefined`, you can use the union type `string | null | undefined`.
```
sandersn: It should be 'want', not 'wanted', right? I am a little confused about the tense here.

mhegazy:
👍

PR #300, opened 2016-06-14
Update compiler options

mhegazy:
Update documentation for compiler options, part of https://github.com/Microsoft/TypeScript-Handbook/issues/294

PR #315, opened 2016-06-23
Fix 291: Add --lib documentation

### LINKED ISSUE #291: Document --lib flag

yuit:
Fix #291


DanielRosenwasser:
@mhegazy already did this in #300. As a heads up, 2.0 specific updates should be pulled into 2.0.

PR #316, opened 2016-06-24
Fix incorrect information about moduleResolution

fay-jai:
Fixes #314 

Based on information in the [module resolution docs](https://www.typescriptlang.org/docs/handbook/module-resolution.html), the default module resolution is now Node and not Classic.


mhegazy:
Thanks for the PR. I beleive I have updated this earlier today in https://github.com/Microsoft/TypeScript-Handbook/commit/30ca8eaea73e2c691a0ed74b950c1d1c17b6807e. take a look and let us know if you think it still needs modifications.

Issue #308, opened 2016-06-22
Add documentation for Shorthand ambient module declarations and wildcard matching in module names


mhegazy:
see title.

andy-ms:
Resolved by #317 and #318

PR #318, opened 2016-06-24
Document wildcard module declarations

andy-ms:
Part of #308

### DIFF
--- a/pages/Module Resolution.md
+++ b/pages/Module Resolution.md
@@ -42,6 +42,33 @@ A non-relative import can be resolved relative to `baseUrl`, or through path map
 They can also resolve to [ambient module declarations](./Modules.md#ambient-modules).
 Use non-relative paths when importing any of your external dependnecies.
 
+## Wildcard module declarations
+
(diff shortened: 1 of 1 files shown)


pages/Module Resolution.md
```diff
@@ -42,6 +42,33 @@ A non-relative import can be resolved relative to `baseUrl`, or through path map
...
+    export default content;
+}
+// Some do it the other way around.
+declare module "json!*" {
+    const value: any;
+    export default value;
+}
+```
+
+Imports of the kind "*!text" and "json!*" can now be used.
```
sandersn: I would (1) change to active voice (2) be more explicit about the type of `"*!text"/"json!*"` ('imports of the kind' sounds vague to me). Something like:

> Now you can import things that match `"*!text"` and `"json!*"`.

Also you should backtick-quote `"*!text"` and `"json!*"` because the stars seem to be parsed as italic right now.

sandersn:
:+1:

mhegazy:
I think this belongs with Modules rather than in Module Resolution.. thoughts?

PR #321, opened 2016-06-24
This parameters

### LINKED ISSUE #310: Add documentation for `this` in functions

sandersn:
1. Document this parameters
2. Document `--noImplicitThis`
3. Merge (1) and (2) with existing discussion of `this` and arrow functions

Fixes #310

### DIFF
--- a/pages/Functions.md
+++ b/pages/Functions.md
@@ -223,7 +223,6 @@ Fortunately, TypeScript lets you catch incorrect uses of `this` with a couple of
 If you need to learn how `this` works in JavaScript, though, first read Yehuda Katz's [Understanding JavaScript Function Invocation and "this"](http://yehudakatz.com/2011/08/11/understanding-javascript-function-invocation-and-this/).
 Yehuda's article explains the inner workings of `this` very well, so we'll just cover the basics here.
 
-
 ## `this` and arrow functions
 
 In JavaScript, `this` is a variable that's set when a function is called.
@@ -303,7 +302,6 @@ function f(this: void) {
 ```
 
 Let's add a couple of interfaces to our example above, `Card` and `Deck`, to make the types clearer and easier to reuse:
-
 
 ```ts
 interface Card {
--- a/pages/Functions.md
+++ b/pages/Functions.md
@@ -289,8 +289,8 @@ It will point out that `this` in `this.suits[pickedSuit]` is of type `any`.
 
 ## `this` parameters
 
-Unfortunately, by default TypeScript will still type `this` in `this.suits[pickedSuit]` as `any`.
-That's because TypeScript doesn't know the type of `this` inside a function expression.
+Unfortunately, the type of `this.suits[pickedSuit]` is still `any`.
+That's because `this` comes from the function expression inside the object literal.
 To fix this, you can provide an explicit `this` parameter.
 `this` parameters are fake parameters that come first in the parameter list of a function:
 
--- a/pages/Functions.md
+++ b/pages/Functions.md
@@ -215,16 +215,20 @@ function buildName(firstName: string, ...restOfName: string[]) {
 let buildNameFun: (fname: string, ...rest: string[]) => string = buildName;
 ```
 
-# Lambdas and using `this`
+# `this`
 
-How `this` works in JavaScript functions is a common theme in programmers coming to JavaScript.
-Indeed, learning how to use it is something of a rite of passage as developers become more accustomed to working in JavaScript.
+Learning how to use `this` in JavaScript is something of a rite of passage.
 Since TypeScript is a superset of JavaScript, TypeScript developers also need to learn how to use `this` and how to spot when it's not being used correctly.
-A whole article could be written on how to use `this` in JavaScript, and many have. Here, we'll focus on some of the basics.
+Fortunately, TypeScript lets you catch incorrect uses of `this` with a couple of techniques.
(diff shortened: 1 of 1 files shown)


pages/Functions.md
```diff
@@ -215,16 +215,19 @@ function buildName(firstName: string, ...restOfName: string[]) {
 let buildNameFun: (fname: string, ...rest: string[]) => string = buildName;
 ```
 
-# Lambdas and using `this`
+# `this`
 
-How `this` works in JavaScript functions is a common theme in programmers coming to JavaScript.
-Indeed, learning how to use it is something of a rite of passage as developers become more accustomed to working in JavaScript.
+Learning how to use `this` in JavaScript is something of a rite of passage.
```
mhegazy: indeed :)

pages/Functions.md
```diff
@@ -279,7 +284,91 @@ let pickedCard = cardPicker();
 alert("card: " + pickedCard.card + " of " + pickedCard.suit);
 ```
 
-For more information on ways to think about `this`, you can read Yehuda Katz's [Understanding JavaScript Function Invocation and "this"](http://yehudakatz.com/2011/08/11/understanding-javascript-function-invocation-and-this/).
+Even better, TypeScript will warn you when you make this mistake if you pass the `--noImplicitThis` flag to the compiler.
+It will point out that `this` in `this.suits[pickedSuit]` is of type `any`.
+
+## `this` parameters
+
+Unfortunately, you will find that TypeScript still says `this` in `this.suits[pickedSuit]` is of type `any` even if you use the arrow function.
```
mhegazy: how about?

By default TypeScriptst will type `this` in `this.suits[pickedSuit]` as `any`

pages/Functions.md
```diff
@@ -279,7 +284,91 @@ let pickedCard = cardPicker();
...
+That's because TypeScript doesn't know the type of `this` inside a function expression.
+To fix this, you can provide an explicit `this` parameter.
+`this` parameters are fake parameters that come first in the parameter list of a function:
+
+```ts
+function f(this: void) {
+    // in strict mode, `this` is undefined for a standalone function,
```
mhegazy: "strict mode" already has a meaning in JS, and it is not something we discussed before, i would rather be explicit and say "with `--noImplictThis` on"

pages/Functions.md
```diff
@@ -279,7 +284,91 @@ let pickedCard = cardPicker();
...
+        // can't use this here because it's of type void!
+        console.log('clicked!');
+    }
+}
+uiElement.addClickListener(Handlers.onClickBad);
+uiElement.addClickListener(Handlers.onClickGood);
+```
+
+Now you will get an error when you try to register `onClickBad` because its `this` is `typeof Handlers`, which `addClickListener` does not accept.
+In contrast, because `onClickGood` specifies its `this` type as `void`, it is legal to pass to `addClickListener`.
```
mhegazy: might be worth to add an explicit pluck-out-and-call example to make the concept clearer..

pages/Functions.md
sandersn: I actually meant JS strict mode, because I'm referring to runtime. Hm, I wonder if I can rewrite this to refer to compile-time only.

sandersn: Changed to 
```ts
    // make sure `this` is unusable in this standalone function
```

mhegazy: oh.. sorry i thought you meant "strict this mode" :) my bad.

pages/Functions.md
sandersn: I expanded it into three parts:

1. bad
2. good
3. good w/arrow function to capture this

pages/Functions.md
```diff
@@ -279,7 +284,116 @@ let pickedCard = cardPicker();
 alert("card: " + pickedCard.card + " of " + pickedCard.suit);
 ```
 
-For more information on ways to think about `this`, you can read Yehuda Katz's [Understanding JavaScript Function Invocation and "this"](http://yehudakatz.com/2011/08/11/understanding-javascript-function-invocation-and-this/).
+Even better, TypeScript will warn you when you make this mistake if you pass the `--noImplicitThis` flag to the compiler.
+It will point out that `this` in `this.suits[pickedSuit]` is of type `any`.
+
+## `this` parameters
+
+Unfortunately, by default TypeScript will still type `this` in `this.suits[pickedSuit]` as `any`.
```
mhegazy: can we make it sound less negative. i mean that there is a default behavior that is there because this is JS, and there is an optional stricter checking mode that users can, and should opt in.

mhegazy:
on a related note, can you file an issue for updating lib.d.ts to use `this` parameters for event handlers.

sandersn:
Where does that issue go? TypeScript or TSJS-lib-generator?

pages/Functions.md
sandersn: Well, the stricter checking of `--noImplicitThis` will just point out the problem -- `--strictThisChecks` did not ship.

mhegazy:
Typescript seems more visible.

pages/Functions.md
sandersn: How about changing this line and the next to:

Unfortunately, the type of `this.suits[pickedSuite]` is still `any`.
That's because `this` comes from the function expression inside the object literal.

sandersn:
done, Microsoft/TypeScript#9374

pages/Functions.md
sandersn: ok, I reworded it to the above.

mhegazy:
👍

PR #323, opened 2016-06-24
Document readonly

### LINKED ISSUE #311: Add documentation readonly properties and index signatures

sandersn:
Fixes #311

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -216,27 +216,42 @@ let howard = new Employee("Howard", "Sales");
 let john = new Person("John"); // Error: The 'Person' constructor is protected
 ```
 
-## Parameter properties
+# Readonly modifier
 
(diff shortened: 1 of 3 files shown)


mhegazy:
can you also add a note about getters with no setters are inferred to be readonly properties.

mhegazy:
👍

sandersn:
I thought about adding that note, but I'm not sure what difference it makes? Just in the .d.ts, right?

mhegazy:
for usage as well. 

```ts
class C {
    get x() { return 0; }
}
c.x = 1; // Error
```

sandersn:
Right, but `readonly` doesn't add anything that wasn't there before. The only change is in the type. I'll make the note.

mhegazy:
might not be worth it... up to you.

sandersn:
I went ahead with it. I didn't think about the `.d.ts` case when I first made the decision not to document it.

mhegazy:
👍

PR #322, opened 2016-06-24
Document UMD module definitions

andy-ms:
Fixes #309

PR #328, opened 2016-06-27
More umd

andy-ms:
Re: https://github.com/Microsoft/TypeScript-Handbook/pull/322#issuecomment-228483753


mhegazy:
can you target release-2.0 instead of master.

PR #329, opened 2016-06-27
Add description of UMD modules to `Modules.md` in addition to `Writin…

andy-ms:
Re: https://github.com/Microsoft/TypeScript-Handbook/pull/322#issuecomment-228483753

### DIFF
--- a/pages/Modules.md
+++ b/pages/Modules.md
@@ -519,7 +519,10 @@ x(y);
 
 ### UMD modules
 
-Some libraries can be used either through imports or through globals.
+Some libraries are designed to be used in many module loaders, or with no module loading (global variables).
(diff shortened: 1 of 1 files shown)


pages/Modules.md
```diff
@@ -517,6 +517,31 @@ import x, {y} from "hot-new-module";
 x(y);
 ```
 
+### UMD modules
+
+Some libraries can be used either through imports or through globals.
```
mhegazy: i would add some elaboration here, and reference that these modules are some times referred to as isomorphic (see http://isomorphic.net/).

RyanCavanaugh:
:+1:

mhegazy:
👍

PR #332, opened 2016-06-27
Add noUnusedLocals and noUnusedParameter

yuit:


mhegazy:
you need to add this to the MSBuild one as well.

yuit:
@mhegazy I thought so but then when I check our msbuild, I don't see them there. Sound like we should add the msbuild there

mhegazy:
👍

PR #333, opened 2016-07-01
Fix typo

andy-ms:


pages/Module Resolution.md
```diff
@@ -407,7 +407,7 @@ Compiling `app.ts` using `--noResolve` should result in:
 
 ## Common Questions
 
-### Why does a module in `exclude` list is still picked up by the compiler
```
DanielRosenwasser: Why does a module in the `exclude` list still get picked up by the compiler.

PR #339, opened 2016-07-13
Fix broken link to tsconfig.json doc

jiayihu:
Fixes #

The link is broken on the [website](http://www.typescriptlang.org/docs/handbook/react-&-webpack.html)


mhegazy:
@DanielRosenwasser does this look like the right fix?

DanielRosenwasser:
No the right fix is to create a series of redirects for the website because these were written with the assumption that they'd be put in `docs/handbook/tutorial/`.

PR #342, opened 2016-07-17
Adds keywords page

### LINKED ISSUE #341: Add a page (or pages) with a list of keywords, and links to documentation
Such a list would make it easier for users to get and retain a fuller grasp of Typescript, by having an overview of the syntax of the language, and by clearly differentiating the parts that are common to ES5/6/7 and the parts that are unique to Typescript.

Presenting this information in this way, would allow users to reduce the question of "What does this keyword do in Typescript?", to "What does this keyword do differently in Typescript than in Javascript?".

Something like the Javascript keyword list on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords) but also including a small one-line description of each keyword, similar to the [Javascript statment list on MSDN](https://msdn.microsoft.com/en-us/library/3xcfcb93(v=vs.94).aspx):

Keyword | Definition
--------|---------------
**const** | [Force the use of const enums](https://www.typescriptlang.org/docs/handbook/enums.html)

For the usages which Typescript shares with Javascript, I'm not sure which would be better -- not to mention them, or to simply include references to authoritative sources of Javascript documentation (MDN and/or MSDN):

Keyword | Language | Definition
---|---|---
**const** | JS | [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const)
 | TS | [Force the use of const enums](https://www.typescriptlang.org/docs/handbook/enums.html)
**var** | JS | [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var) [MSDN] (https://msdn.microsoft.com/en-us/library/z16cackw(v=vs.94).aspx)

zspitz:
Fixes #341 

cc @mhegazy


zspitz:
Closing this so I can retarget `release-2.0` via #344

PR #344, opened 2016-07-17
Create Keywords.md

### LINKED ISSUE #341: Add a page (or pages) with a list of keywords, and links to documentation

zspitz:
Fixes #341 

cc @mhegazy


zspitz:
Microsoft/Typescript#2536

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+Unless otherwise noted, more information can be found on the [Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Web/JavaScript) or on [MSDN](https://msdn.microsoft.com/en-us/library/d1et7k7c%28v=vs.94%29.aspx)
+
+---
+
+The following keywords are used for [basic type annotations](http://www.typescriptlang.org/docs/handbook/basic-types.html):
+
+Keyword | Description
+---|---
+any | [Describes a type unknown at design time](http://www.typescriptlang.org/docs/handbook/basic-types.html#any)
+boolean | [Boolean](http://www.typescriptlang.org/docs/handbook/basic-types.html#boolean)
```
mhegazy: use lower case type names, so as not to mix them with boxed types, Number, String, etc..

mhegazy: also add `never`.

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+null | _Missing Handbook link_ ([Spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#326-the-null-type))
+number | [Number](http://www.typescriptlang.org/docs/handbook/basic-types.html#number)
+string | [String](http://www.typescriptlang.org/docs/handbook/basic-types.html#string)
+symbol | [Symbol](http://www.typescriptlang.org/docs/handbook/symbols.html)
+undefined | _Missing Handbook link_ ([Spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#327-the-undefined-type))
+void | _Missing Handbook link_ ([Spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#325-the-void-type))
```
mhegazy: we should also add `this`. as it has a special meaning.

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+
+---
+
+The following keywords are used for [basic type annotations](http://www.typescriptlang.org/docs/handbook/basic-types.html):
+
+Keyword | Description
+---|---
+any | [Describes a type unknown at design time](http://www.typescriptlang.org/docs/handbook/basic-types.html#any)
+boolean | [Boolean](http://www.typescriptlang.org/docs/handbook/basic-types.html#boolean)
+null | _Missing Handbook link_ ([Spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#326-the-null-type))
```
mhegazy: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Basic%20Types.md#null-and-undefined

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+number | [Number](http://www.typescriptlang.org/docs/handbook/basic-types.html#number)
```
mhegazy: make links relative to this one. the contents of the handbook should be navigatable as md files on github. the script the publishes them to the website takes care of fixing the links.

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+string | [String](http://www.typescriptlang.org/docs/handbook/basic-types.html#string)
+symbol | [Symbol](http://www.typescriptlang.org/docs/handbook/symbols.html)
+undefined | _Missing Handbook link_ ([Spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#327-the-undefined-type))
```
mhegazy: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Basic%20Types.md#null-and-undefined

pages/Keywords.md
mhegazy: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Basic%20Types.md#void

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+void | _Missing Handbook link_ ([Spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#325-the-void-type))
+
+Other basic type-related keywords:
+
+Keyword | Description
+---|---
+as | [Type assertion](http://www.typescriptlang.org/docs/handbook/basic-types.html#type-assetions)
```
mhegazy: also used as a module import rename declaration.

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+is | _Missing link_
```
mhegazy: see https://github.com/Microsoft/TypeScript-Handbook/blob/28108e225b85aa29666256c0db9a4a61b271bbe8/pages/Advanced%20Types.md#user-defined-type-guards

pages/Keywords.md
```diff
@@ -0,0 +1,97 @@
...
+
+User-defined types:
+
+Keyword | Description
+---|---
+enum | [Defines an enum - a set of named values](http://www.typescriptlang.org/docs/handbook/enums.html)
+type | [Type alias](http://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases)
+interface | [Defines a type by its shape (structural typing)](http://www.typescriptlang.org/docs/handbook/interfaces.html)
```
mhegazy: why not class too?

mhegazy:
thanks @zspitz. since this is intended to be an index of sorts, i would recommend merging them all in one table. left column  has one key word, all sorted alphabetically of course. the right column can have multiple links based on the context. for the JS keywords, i would also add links to MDN documentation or ES6 spec.

zspitz:
@mhegazy I'm thinking that an alphabetical arrangement might be important on paper, but within the browser there's not much benefit -- it's always possible to do CTRL+F to find a specific keyword on the page. However, isn't there added value in the grouping - type annotations, type modifiers, member modifiers etc.?
(I'll take care of the rest shortly.)

mhegazy:
Well.. then what is the value of this page in general. you can always search this repo for keywords and that should get you close. like `class` would get you all the relevant areas. I was thinking of this more of a keyword index to get you quickly where you want, so you can look up `as`, and know it is used in 1. ES6 module declaration or 2. a post fix type assertion operator; similarly `async` would get you to some documentation about async functions in general.

zspitz:
@mhegazy 
> you can always search this repo for keywords and that should get you close. like class would get you all the relevant areas.

I think many people would be coming from the Handbook link on the Typescript website, and not through the repo, thus repo search can't be counted on. Unless the plan is to eventually have that link refer to the repo (similar to the spec), and not to http://www.typescriptlang.org/docs?

> so you can look up `as`, and know it is used in ...
True. But perhaps sometimes you want to look up the available usages (basic type, type modifier, property modifier etc.) and quickly find out from there which keywords can fill each role. (I'm thinking computed enums here...). Admittedly, the hard part will be to figure out the different usage groups first.

pages/Keywords.md
zspitz: Fixed `void`

>`this` as it has a special meaning

Does Typescript add more than the Javascript meaning of global context / call dependent context?

pages/Keywords.md
zspitz: Fixed lowercase and `never`

pages/Keywords.md
zspitz: Fixed

pages/Keywords.md
zspitz: Fixed

pages/Keywords.md
zspitz: Fixed

pages/Keywords.md
zspitz: Fixed

pages/Keywords.md
```diff
@@ -0,0 +1,99 @@
...
+namespace | [Associates the contained types with the specified namespace](namespaces.md)
+require | [Missing description](modules.md#export--and-import--require)
+
+---
+
+Unknown:
+
+Keyword |
+---|
+of|
```
DanielRosenwasser: `for-of`

zspitz: @DanielRosenwasser fixed

pages/Keywords.md
zspitz: Fixed

zspitz:
@mhegazy @DanielRosenwasser I've reorganized the keyword list in simple alphabetical order, and added links as appropriate (Handbook `||` spec `||` MDN)

1. The _this_-type referred to [here](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#363-this-types) isn't a keyword; it refers to the type, but the `this` in code refers to the instance of the class. Other than that, `this` by itself has no added meaning in Typescript over Javascript; except for this-based type guards, which I noted.

2. There's an inconsistency between `if...else` which doesn't have a separate entry for `else`, and `as`, `from`, and `extends` which even though they can't stand alone, each have their own entry. The difference is between Javascript constructs and Typescript constructs.

zspitz:
@mhegazy  @DanielRosenwasser I think this is pretty much complete. Could you look over it and see if I've missed anything? Any other thoughts?

pages/Keywords.md
```diff
@@ -0,0 +1,173 @@
...
+| `as` | | [Type operation](#type-operation) | [Type assertion](Basic%20Types.md#type-assertions) | |
+| `as` | `import {`_item_ `as` _name_`} from "`_path_`"`<br>`import * as `_name_` from "` _path_ `"` | [Environment/modules](#environment-modules) | [Module import renaming](Modules.md#import) | |
+| `as` | `export as namespace` _namespace_ | [Environment/modules](#environment-modules) | [UMD/isomorphic modules](Modules.md#umd-modules) | |
+| `async` | | [Functions](#functions) | Marks function as asynchronous | [ES draft](http://tc39.github.io/ecmascript-asyncawait/), [PR#1664](https://github.com/Microsoft/TypeScript/issues/1664), [spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#68-asynchronous-functions) |
+| `await` | | [Functions](#functions) | Waits for value within an asynchronous function | [ES draft](http://tc39.github.io/ecmascript-asyncawait/), [PR#1664](https://github.com/Microsoft/TypeScript/issues/1664), [spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#68-asynchronous-functions) |
+| `boolean` | | [Predefined type](#predefined-types) | [`boolean` type](Basic%20Types.md#boolean) | |
```
mhegazy: use `"Primitive Type"` instead of `"Predefined type"`. this is how the TypeScript spec refers to them anyways.

mhegazy: this also applies to the rest of the document.

mhegazy:
sorry for the delay, have been out on vacation for a while. a few comments:

- use "Primitive Type" instead of "Predefined type"
- do not think the syntax column is adding much value, it is not filled for most entries, i would just remove it.
- In the first table, description and category are mislabeled. seems like he description column has the category and the category column has the description .

zspitz:
@mhegazy Fixed 'primitive type' to 'predefined type', and headings.
RE: syntax column -- When the same keyword has different meanings (like `abstract` which is applied to both classes and members), the syntax column clarifies the difference between them. This is particularly true when the meanings have little or no relation to each other (e.g. the distinction between `this` function parameters, and `this`-based type guards).
I had originally thought to have the syntax in the first column, but I think the order becomes confusing when the syntax doesn't start with the keyword (`default` vs. `export default` _expr_).

metaleap:
Status on this?  =)

zspitz:
@mhegazy @metaleap There are some changes that need to be made here for 2.1 and 2.2. But if the `release-2.0` branch is obsolete, and this PR needs to be merged somehow into the master branch, then I guess that should be done first.

RyanCavanaugh:

----

Saved reply above; this technically doesn't have merge conflicts but has a ton of unrelated changes in one PR and I'm not comfortable merging it.

PR #349, opened 2016-07-19
Add tutorial on migrating from JavaScript

DanielRosenwasser:
@bowdenk7 @mhegazy @sandersn @RyanCavanaugh any feedback you have would be valuable here.

### DIFF
--- a/pages/tutorials/Migrating from JavaScript.md
+++ b/pages/tutorials/Migrating from JavaScript.md
@@ -1,0 +1,333 @@
+TypeScript doesn't exist in a vacuum.
+It was built with the JavaScript ecosystem in mind, and a lot of JavaScript exists today.
+Converting a JavaScript codebase over to TypeScript is, while tedious, typically not challenging.
+In this tutorial, we're going to look at how you might start out.
+
+# Setting up your Directories
+
+If you're writing in plain JavaScript, it's likely that you might be running your JavaScript directly.
+That means you might be writing `.js` files in a `src`, `lib`, or `dist` directory, and then running it however you've needed.
+
+If that's the case, the files that you've written are going to used as inputs to TypeScript, and you'll run the outputs it produces.
+While you can do this in the same folder, this doesn't work if you have any `.js` files because those input files would be overwritten with the outputs.
+Instead, we'll want to separate these inputs and outputs between different folders.
+We recommend calling your source (input) directory `src`, and your output directory `lib` or `built`.
+If you need to preserve the original directory name for some reason, feel free to play around with these.
+
+You might also be running some intermediate steps on your JavaScript, such as bundling or using another transpiler like Babel.
+In this case, you might already have a folder structure like this set up.
+
+From this point on, we're going to assume that your directory is set up something like this:
+
+```text
+projectRoot
+├── src
+│   ├── file1.js
+│   └── file2.js
+├── built
+└── tsconfig.json
+```
+
+You might have a `tests` folder which might be under a separate folder from your main JavaScript files.
+In that case, you might have one `tsconfig.json` in `tests`, and one in `tests` as well, and apply the rest of this tutorial to that setup.
+
+# Writing a Configuration File
+
+TypeScript uses a file called `tsconfig.json` for managing your project's options, such as which files you want to include, and what sorts of checking you want to perform.
+Let's create a bare-bones one for our project:
+
+```json
+{
+    "compilerOptions": {
+        "outDir": "./built",
+        "allowJS": true,
+        "target": "es5"
+    },
+    "include": [
+        "./src"
+    ]
+}
+```
+
+Here we're specifying a few things to Typescript:
+
+1. To read in any files it understands in the `src` directory (with `include`).
+2. To accept JavaScript files as inputs (with `allowJs`).
+3. To emit all of the output files in `built` (with `outDir`).
+4. Translate newer JavaScript constructs down to an older version like ECMAScript 5 (using `target`).
+
+At this point, if you try running `tsc` at the root of your project, you should see output files in the `built` directory.
+The layout of files in `built` should look identical to the layout of `src`.
+You should now have TypeScript working with your project.
+
+## Early Benefits
+
+Even from this point on, you can get some great benefits from TypeScript understanding your project.
+If you open up an editor like Visual Studio Code or Visual Studio, you'll see that you can often get some tooling support like completion.
+You can also catch certain bugs with options like:
+
+* `noImplicitReturns` which prevents you from forgetting to return at the end of a function.
+* `noFallthroughCasesInSwitch` which is helpful if you never want to forget a `break` statement between `case`s in a `switch` block.
+
+TypeScript will also warn about unreachable code and labels, which you can disable with `allowUnreachableCode` and `allowUnusedLabels` respectively.
+
+# Integrating with Build Tools
+
+You might have some more build steps in your pipeline.
+Perhaps you concatenate something to each of your files.
+Each build tool is different, but we'll do our best to cover the gist of things.
+
+## Gulp
+
+If you're using Gulp in some fashion, we have a tutorial on [using Gulp](./Gulp.md) with TypeScript, and integrating with common build tools like Browserify, Babelify, and Uglify.
+You can read more there.
+
+## Webpack
+
+Webpack integration is pretty simple.
+You can use `ts-loader`, a TypeScript loader, combined with `source-map-loader` for easier debugging.
+Simply run
+
+```shell
+npm install ts-loader source-map-loader
+```
+
+and merge in options from the following into your `webpack.config.js` file:
+
+```js
+module.exports = {
+    entry: "./src/index.ts",
+    output: {
+        filename: "./dist/bundle.js",
+    },
+
+    // Enable sourcemaps for debugging webpack's output.
+    devtool: "source-map",
+
+    resolve: {
+        // Add '.ts' and '.tsx' as resolvable extensions.
+        extensions: ["", ".webpack.js", ".web.js", ".ts", ".tsx", ".js"]
+    },
+
+    module: {
+        loaders: [
+            // All files with a '.ts' or '.tsx' extension will be handled by 'ts-loader'.
+            { test: /\.tsx?$/, loader: "ts-loader" }
+        ],
+
+        preLoaders: [
+            // All output '.js' files will have any sourcemaps re-processed by 'source-map-loader'.
+            { test: /\.js$/, loader: "source-map-loader" }
+        ]
+    },
+
+    // Other options...
+};
+```
+
+It's important to note that `ts-loader` will need to run before any other loader that deals with `.js` files.
+
+# Moving to TypeScript
+
+At this point, you're probably ready to start using TypeScript files.
+The first step is to rename one of your `.js` files to `.ts`.
+If your file uses JSX, you'll need to rename it to `.tsx`.
+
+Finished with that step?
+Great!
+You've successfully migrated a file from JavaScript to TypeScript!
+
+Of course, that might not feel right.
+If you open that file in an editor with TypeScript support (or if you run `tsc --pretty`), you might see red squiggles on certain lines.
+You should think of these the same way you'd think of red squiggles in an editor like Microsoft Word.
+TypeScript will still translate your code, just like Word will still let you print your documents.
+
+That might not be the best behavior for now, but you can think of this process as a knob on the type of strictness you'd like to work with.
+If, for instance, you *don't* want TypeScript to compile to JavaScript in the face of errors, you can use the `noEmitOnError` option.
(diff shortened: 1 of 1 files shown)


pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
+TypeScript doesn't exist in a vacuum.
+It was built with the JavaScript ecosystem in mind, and a lot of JavaScript exists today.
+Converting a JavaScript codebase over to TypeScript is, while tedious, typically not challenging.
+In this tutorial, we're going to look at how you might start out.
+
+# Setting up your Directories
+
```
sandersn: This section meanders too much. I'd start off with a statement of the problem and give the proposed folder structure first, then compare with the existing vanilla or Babel setup. Something like:

Typescript takes input source files and produces output JavaScript. To avoid overwriting your JavaScript source, the best folder structure is: *folder diagram here*. 

This reads from source files from src and outputs files to built. Then you run the files in built. You may already have a similar layout if you're bundling or using another transpiler like Babel. This setup keeps source JavaScript from being overwritten by output with the same file name, and it keeps you from opening the output files by mistake.

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+```text
+projectRoot
+├── src
+│   ├── file1.js
+│   └── file2.js
+├── built
+└── tsconfig.json
+```
+
+You might have a `tests` folder which might be under a separate folder from your main JavaScript files.
```
sandersn: Cut one of the 'might's: You might have a tests folder separate from your main JavaScript files.

sandersn: And the next line would be "In that case, you would have ...".

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+3. To emit all of the output files in `built` (with `outDir`).
+4. Translate newer JavaScript constructs down to an older version like ECMAScript 5 (using `target`).
+
+At this point, if you try running `tsc` at the root of your project, you should see output files in the `built` directory.
+The layout of files in `built` should look identical to the layout of `src`.
+You should now have TypeScript working with your project.
+
+## Early Benefits
+
+Even from this point on, you can get some great benefits from TypeScript understanding your project.
```
sandersn: "Even from this point on" &rarr; "Even at this point"

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+If you open up an editor like Visual Studio Code or Visual Studio, you'll see that you can often get some tooling support like completion.
```
sandersn: We should have links to Code and VS.

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+The first checks states that `myCoolFunction` takes a function (which takes a `number`), and then a list of `number`s.
+The second one says that it will take a function as well, and then uses a rest parameter (`...nums`) to state that any number of arguments after that need to be `number`s.
+
+### `any`, `Object`, and `{}`
+
+You might be tempted to use `Object` or `{}` to say that a value can have any property on it, because `Object` is, for most purposes, the most general type.
+But **`any` is actually the type you want to use** in those situations, since it's the most flexible type.
+
+For instance, if you have something that's typed as `Object` you won't be able to call methods like `toLowerCase()` on it.
+`any` is simulataneously the most general and most capable type, and will allow you to call it, construct it, access properties on it, etc.
```
sandersn: Typo: simultaneously

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Strict `null` & `undefined` Checks
+
+By default, TypeScript assumes that `null` and `undefined` are in the domain of every type.
+That means anything declared with the type `number` could be `null` or `undefined`.
+Since `null` and `undefined` are such a frequent source of bugs in JavaScript and TypeScript, TypeScript has the `strictNullChecks` option.
+
+When `strictNullChecks` is enabled, `null` and `undefined` get their own types.
+Whenever anything is *possibly* `null`, you could use a union type with the original type.
+So for instance, if something could be a `number` or `null`, you'd write the type out as `number | null`.
```
sandersn: Practically, you should also note that this doesn't work with all libraries yet.

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+You should now have TypeScript working with your project.
+
+## Early Benefits
+
+Even from this point on, you can get some great benefits from TypeScript understanding your project.
+If you open up an editor like Visual Studio Code or Visual Studio, you'll see that you can often get some tooling support like completion.
+You can also catch certain bugs with options like:
+
+* `noImplicitReturns` which prevents you from forgetting to return at the end of a function.
+* `noFallthroughCasesInSwitch` which is helpful if you never want to forget a `break` statement between `case`s in a `switch` block.
```
sandersn: That works with --allowJs? Very cool. I thought we didn't report errors on JS files. I think I'm confusing allowJs with Salsa though.

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+            // All output '.js' files will have any sourcemaps re-processed by 'source-map-loader'.
+            { test: /\.js$/, loader: "source-map-loader" }
+        ]
+    },
+
+    // Other options...
+};
+```
+
+It's important to note that `ts-loader` will need to run before any other loader that deals with `.js` files.
```
sandersn: Shouldn't you link to the we pack tutorial at the end here?

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+
+# Moving to TypeScript
+
+At this point, you're probably ready to start using TypeScript files.
```
sandersn: Get rid of 'probably'. The reader is *totally* ready. But we don't need to remind them.

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+You've successfully migrated a file from JavaScript to TypeScript!
+
+Of course, that might not feel right.
+If you open that file in an editor with TypeScript support (or if you run `tsc --pretty`), you might see red squiggles on certain lines.
+You should think of these the same way you'd think of red squiggles in an editor like Microsoft Word.
+TypeScript will still translate your code, just like Word will still let you print your documents.
+
+That might not be the best behavior for now, but you can think of this process as a knob on the type of strictness you'd like to work with.
+If, for instance, you *don't* want TypeScript to compile to JavaScript in the face of errors, you can use the `noEmitOnError` option.
+Eventually you can turn that knob up to 11/10 if you need.
```
sandersn: 1. Switch this sentence with the previous one. `--noEmitError` is an example of the knob set to 11.
2. I would just make it 11. Or even 'eleven'.

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Weeding out Errors
+
```
sandersn: I'm kind of allergic to /not un.*/. How about "Like we mentioned, don't be surprised if you get errors after conversion."

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -331,3 +330,39 @@ Since `null` and `undefined` are such a frequent source of bugs in JavaScript an
...
+}
+```
+
+This has the same problems we mentioned above - we could easily have misspelled `getDistance` and not gotten an error.
+For this reason, TypeScript has the `noImplicitThis` option.
+When that option is set, TypeScript will issue an error when `this` is used without an explicit (or inferred) type.
+The fix is to use a `this`-parameter to give an explicit type in the interface or in the function itself:
+
+```ts
+Point.prototype.distanceFromOrigin = function(this: Point, point: Point) {
```
sandersn: This is a weird one because we actually infer `this` for the method pattern in Salsa, but not in Typescript proper. I guess it makes sense that you will have to add annotations when you switch to Typescript though.

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,368 @@
...
+}
+```
+
+We added two overload signatures to `myCoolFunction`.
+The first checks states that `myCoolFunction` takes a function (which takes a `number`), and then a list of `number`s.
+The second one says that it will take a function as well, and then uses a rest parameter (`...nums`) to state that any number of arguments after that need to be `number`s.
+
+### `any`, `Object`, and `{}`
+
+You might be tempted to use `Object` or `{}` to say that a value can have any property on it, because `Object` is, for most purposes, the most general type.
```
sandersn: Do you think you should be a little more negative in this section? `any` is typically the choice when you don't know the type or can't express it in Typescript. And its flexibility and capability comes at the cost of the tools knowing nothing about the type. Plus it's kind of infectious.

pages/tutorials/Migrating from JavaScript.md
DanielRosenwasser: Ugh apparently not? I could have sworn I'd seen it working before. =/

pages/tutorials/Migrating from JavaScript.md
DanielRosenwasser: How can you be so certain? :smiley:

pages/tutorials/Migrating from JavaScript.md
DanielRosenwasser: Are you non-sarcastically asking me to be more negative? That's actually a first.

DanielRosenwasser:
Updated

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,416 @@
...
+Alternatively, you can just say `options` has the type `any` which is the easiest thing to do, but which will benefit you the least.
+
+### `any`, `Object`, and `{}`
+
+You might be tempted to use `Object` or `{}` to say that a value can have any property on it because `Object` is, for most purposes, the most general type.
+But **`any` is actually the type you want to use** in those situations, since it's the most *flexible* type.
+
+For instance, if you have something that's typed as `Object` you won't be able to call methods like `toLowerCase()` on it.
+Being more general means being less capable.
+`any` is special in that it's simultaneously the most general and most capable type, and will allow you to call it, construct it, access properties on it, etc.
```
sandersn: >most [](start = 66, length = 4)

Should you change "most capable" to "least capable" here? I think you still need to reconcile the new previous line and this line better.

DanielRosenwasser: What about

```text
Being more general usually means you can do less with a type, but `any` is special in that it is the most general type while also being the most capable.
That means you can call it, construct it, access properties on it, etc.
```

sandersn: Sounds good, but you should still mention that it also opts out of editor support and checking, and tends to spread throughout your program.

---
In reply to: [73719584](https://github.com/Microsoft/TypeScript-Handbook/pull/349#discussion_r73719584) [](ancestors = 73719584)

pages/tutorials/Migrating from JavaScript.md
```diff
@@ -0,0 +1,416 @@
+TypeScript doesn't exist in a vacuum.
+It was built with the JavaScript ecosystem in mind, and a lot of JavaScript exists today.
+Converting a JavaScript codebase over to TypeScript is, while somewhat tedious, usually not challenging.
+In this tutorial, we're going to look at how you might start out.
+We assume you've read enough of the handbook to write new TypeScript code.
+
+# Setting up your Directories
+
+If you're writing in plain JavaScript, it's likely that you're running your JavaScript directly,
+  where your `.js` files in a `src`, `lib`, or `dist` directory, and then ran as desired.
```
sandersn: >   [](start = 0, length = 2)

join line 10 to line 9

DanielRosenwasser: @RyanCavanaugh convinced me this is a decently useful pattern to split up a line. No bueno?

sandersn:
:clock1:

pages/tutorials/Migrating from JavaScript.md
sandersn: Oh, if we're going to change our style guide, that's fine. I don't care one way or another as long as we're consistent.

---
In reply to: [73721096](https://github.com/Microsoft/TypeScript-Handbook/pull/349#discussion_r73721096) [](ancestors = 73721096)

PR #351, opened 2016-07-21
Add instructions on how to read the handbook.

jish:
I'm getting started with TypeScript and I was linked to this
repository from a few different places. I do not believe the readme is
informative enough to newcomers.

After some more research it seems that this handbook is meant to be
read online, or from the Markdown files directly in this repository.

If this is incorrect, I will gladly update the readme, or I would
appreciate it if someone more knowledgeable could provide some insight
thanks :)


DanielRosenwasser:
Thanks @jish!

PR #356, opened 2016-08-07
Fix grammatical error

Croccifixio:
Fixes #
The context of the change is presented below:
> Shadowing should *usually* be avoided in the interest of ~~write~~ writing clearer code.


DanielRosenwasser:
Thanks for the fix! Should be out the next time we publish.

Issue #371, opened 2016-08-26
ASP.NET Core tutorial: Typings -- main.d.ts vs. index.d.ts


jlost:
After the part where you import typings, the text on the page tells you to add a reference to `typings/index.d.ts` to your `tsconfig.json.` However, the code showing the actual contents of `tsconfig.json` (on the page) has `typings/main.d.ts` instead. The latter filename (`main.d.ts`) is incorrect and doesn't exist. Maybe it was accurate in a previous version of typings.

Sorry, I'd already created this issue before I noticed the documentation was wiki-based: https://github.com/Microsoft/TypeScript/issues/10519

PR #372, opened 2016-08-26
Fixed typings/index.d.ts filename in tsconfig.json

jlost:
Corrected `typings/main.d.ts` to `typings/index.d.ts` for #371. `index` is what is stated in the above paragraph and is the correct filename (`main.d.ts` doesn't exist in the `typings` folder).


DanielRosenwasser:
Thank you!

DanielRosenwasser:
As a heads up, we'll need to sync this up with the website soon, but probably won't go out today.

PR #375, opened 2016-09-01
Wrap `string` in backticks

jessepinho:


mhegazy:
thanks. applied manually.

PR #376, opened 2016-09-01
Fix grammatical errors

jessepinho:
Before these corrections, both sentences incorrectly used [comma splices](https://en.wikipedia.org/wiki/Comma_splice).


mhegazy:
thanks. applied manually due to conflicts.

PR #381, opened 2016-09-12
Basic Types: refine template string code sample.

weakish:
Replace unnecessary usage of template string with regular string.

---
 pages/Basic Types.md | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -39,7 +39,7 @@ You can also use *template strings*, which can span multiple lines and have embe
 These strings are surrounded by the backtick/backquote (`` ` ``) character, and embedded expressions are of the form `${ expr }`.
 
 ```ts
-let fullName: string = `Bob Bobbington`;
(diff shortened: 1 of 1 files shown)


mhegazy:
why is it unnecessary? template strings are allowed to have no substitution spans, and this can be used in place of any string literal.

mhegazy:
this example is intentional. so do not think this PR is needed.

weakish:
@mhegazy  If it is intentional, I recommend comment it explicitly. 
Using template string to represent normal string is not preferred, since it gives reader false expectation
(a reader may think there is some interpolation  within the string). Thus it wastes reader's time if the string is long.

mhegazy:
> Using template string to represent normal string is not preferred, since it gives reader false expectation

Why? it is an ES6 feature, and a valid JS literal expression, see http://www.ecma-international.org/ecma-262/6.0/#sec-template-literals

PR #383, opened 2016-09-15
Add documentation for the `never` type

mhegazy:
Fixes https://github.com/Microsoft/TypeScript-Handbook/issues/313


pages/Basic Types.md
```diff
@@ -211,6 +211,34 @@ Once again, more on union types later on.
 
 > As a note: we encourage the use of `--strictNullChecks` when possible, but for the purposes of this handbook, we will assume it is turned off.
 
+# Never
+
+The `never` type represents the type of values that never occur.
+For instance, `never` is the return type for a function that always throws an exception or one that never returns;
```
DanielRosenwasser: Is that true? Do we infer that? Because I'm pretty sure we still infer `void`.

pages/Basic Types.md
```diff
@@ -211,6 +211,34 @@ Once again, more on union types later on.
...
+`never` is also the type of variables under type guards that are never true.
```
DanielRosenwasser: Variables also acquire the type `never` when narrowed by any type guards that can never be true.

pages/Basic Types.md
```diff
@@ -211,6 +211,34 @@ Once again, more on union types later on.
...
+
+The `never` type is a subtype of and assignable to every type; however, no type is a subtype of or assignable to `never` (except `never` itself).
```
DanielRosenwasser: This includes the `any` type.

pages/Basic Types.md
mhegazy: well.. only for function expressions and lambda's. i should be more specific.

pages/Basic Types.md
```diff
@@ -211,6 +211,35 @@ Once again, more on union types later on.
 
 > As a note: we encourage the use of `--strictNullChecks` when possible, but for the purposes of this handbook, we will assume it is turned off.
 
+# Never
+
+The `never` type represents the type of values that never occur.
+For instance, `never` is the return type for a function expression or a arrow function expresssion that always throws an exception or one that never returns;
```
chicoxyzzy: expresssion -> expression

PR #390, opened 2016-09-23
Fix for React & Webpack md

polizz:
Fixes #

It is late and I am tired, so I may not be seeing this correctly, but I believe you meant to say the Hello component is being relatively referred to.


mhegazy:
thanks, applied the change manually.

PR #391, opened 2016-09-23
Update Modules.md: an module -> a module

Radagaisus:
Fixed a typo in `modules.md`: an module -> a module.


mhegazy:
Applied manually.

PR #392, opened 2016-09-23
fixed bug in configureation example

hfinger:
The example configuration was not working (i.e. error TS5023: Unknown compiler option 'allowJS'.). This fixes the documentation with lower case letter "s".


mhegazy:
Should be fixed now. thanks.

PR #397, opened 2016-09-29
Fix tuple wording

cironunes:


mhegazy:
this looks grammatically correct to me.

PR #404, opened 2016-10-04
fix options in the example tsconfig.json file

marzelin:
Fixes #

Since only 'amd' and 'system' can be used in conjunction with --outFile, the `module` option was changed from `commonjs` to `system`


DanielRosenwasser:
Could you instead just remove `outFile` since CommonJS is the most common module target?

mhegazy:
I have removed one of them as part of merging this change.

PR #405, opened 2016-10-04
fix the example webpack configuration file

marzelin:
Fixes #

According to webpack documentation output.filename is used solely for naming the individual files. The output.path option (which is required) determines the location on disk where the files are written to.

### DIFF
--- a/pages/tutorials/React & Webpack.md
+++ b/pages/tutorials/React & Webpack.md
@@ -183,7 +183,8 @@ Create a `webpack.config.js` file at the root of the project directory.
 module.exports = {
     entry: "./src/index.tsx",
     output: {
-        filename: "./dist/bundle.js",
+        filename: "bundle.js",
+        path: "./dist"
     },
 
     // Enable sourcemaps for debugging webpack's output.
--- a/pages/tutorials/React & Webpack.md
+++ b/pages/tutorials/React & Webpack.md
@@ -184,7 +184,7 @@ module.exports = {
     entry: "./src/index.tsx",
     output: {
         filename: "bundle.js",
-        path: "./dist"
+        path: __dirname + "/dist"
     },
 
     // Enable sourcemaps for debugging webpack's output.


pages/tutorials/React & Webpack.md
```diff
@@ -183,7 +183,8 @@ Create a `webpack.config.js` file at the root of the project directory.
 module.exports = {
     entry: "./src/index.tsx",
```
TheLarkInn: If we are going to fix it we should probably use the verbose `entry` syntax: 

```
entry: {
  app: './src/index.tsx'
}
```

pages/tutorials/React & Webpack.md
```diff
@@ -183,7 +183,8 @@ Create a `webpack.config.js` file at the root of the project directory.
...
     output: {
-        filename: "./dist/bundle.js",
+        filename: "bundle.js",
+        path: "./dist"
```
TheLarkInn: this needs to be an absolute path. 

webpack 2 will error for this, webpack 1 will not, but you should be specifying absolute path.

TheLarkInn:
cc @DanielRosenwasser

pages/tutorials/React & Webpack.md
marzelin: There's only one entry, so I don't see any reason for using syntax for multiple entries

pages/tutorials/React & Webpack.md
marzelin: Ok, I'll update the code so that it's ready for webpack 2.

pages/tutorials/React & Webpack.md
TheLarkInn: Just a best practice recommendation. In the end the short hand you have gets converted to that behind the scenes.

TheLarkInn: http://webpack.js.org/concepts/entry-points/ If you want reference.

marzelin: Simplicity is important especially in a tutorial, but shouldn't we follow good practices in the first place? The current config in invalid and works just because of sheer luck.

DanielRosenwasser: Is it invalid @TheLarkInn?

TheLarkInn: Looking just at the config the entry property appears valid.

marzelin: @TheLarkInn I also think that `entry` part is valid. This *pr* is about changing `output` part. @DanielRosenwasser stated in the comments above that he don't want to change the `output` path to keep tutorial simpler, but I think it should be changed because the current `output` is no longer valid according to webpack documentation. As a webpack core team member can you please comment on this subject?

TheLarkInn: Oh yeah, definitely want to have the filename and path separated like that. It is by fluke that that actually works.

marzelin: @TheLarkInn thank you for the review. I think we are ready to merge this pr @DanielRosenwasser

DanielRosenwasser: Thanks for clarifying - sorry if this came off as bikeshedding. 😄

DanielRosenwasser:
Thanks @marzelin!

Issue #406, opened 2016-10-05
types in tsconfig.json


gertcuykens:
https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/tsconfig.json.md
The tsconfig manual doesn't mention anything about the `types` propertie? For example { "types" : [ "es2015", "dom", "customElements" } }

https://github.com/Polymer/polymer/pull/3954/files
![image](https://cloud.githubusercontent.com/assets/863731/19107349/0f71f502-8aec-11e6-9985-94b3c6555673.png)

I am I looking at the wrong place?

PR #407, opened 2016-10-05
Add documentation for `types` and `typeRoots` in tsconfig.json

mhegazy:
Fixes https://github.com/Microsoft/TypeScript-Handbook/issues/406


mhegazy:
//cc @RyanCavanaugh and @DanielRosenwasser

pages/tsconfig.json.md
```diff
@@ -95,6 +95,42 @@ A `tsconfig.json` file is permitted to be completely empty, which compiles all f
 
 Compiler options specified on the command line override those specified in the `tsconfig.json` file.
 
+## `@types`, `typeRoots` and `types`
+
+By default all *visible* "`@types`" packages are included in your compilation.
+Packages in `node_modules/@types` of any enclosing folder are considered *visible*;
+specifically, that means packages within `./node_modules/@types/`,  `../node_modules/@types/`, `../../node_modules/@types/`, and so on.
```
DanielRosenwasser: Is this true? My understanding is that these paths are only used when *resolving*. Only the cwd's `node_modules/@types` is included.

mhegazy: nope. we made a change in 2.0.3 to allow them all after getting feedback that the current working directory one is confusing.

DanielRosenwasser: Yay!

pages/tsconfig.json.md
```diff
@@ -95,6 +95,41 @@ A `tsconfig.json` file is permitted to be completely empty, which compiles all f
...
+   "compilerOptions": {
+       "types" : ["node", "lodash", "express"]
+   }
+}
+```
+
+This `tsconfig.json` file will *only* include  `./node_modules/@types/node`, `./node_modules/@types/lodash` and `./node_modules/@types/express`.
+Other packages under `node_modules/@types/*` will not be included.
+
+Specify `"types": []` to disable automatic inclusion of `@types` packages.
```
DanielRosenwasser: It might be worth it to state something like

```md
Keep in mind that automatic inclusion is only important if you're writing global code (as opposed to code that uses modules.
If you use an `import "foo"` statement, for instance, TypeScript may still look through `node_modules` & `node_modules/@types` folders to find the `foo` package.

Comment (DanielRosenwasser) [inline pages/tsconfig.json.md]: It might be worth it to state something like

```md
Keep in mind that automatic inclusion is only important if you're writing global code (as opposed to code that uses modules.
If you use an `import "foo"` statement, for instance, TypeScript may still look through `node_modules` & `node_modules/@types` folders to find the `foo` package.
```

mhegazy: the issue here is it depends on how the @types package is authored
you could do something like `import ... from "bluebird"` from your perspective this is a module, though the `@types` package define it as a global, so you need to add the `types` entry for it.

gertcuykens:
For me it's also confusing when I should use `lib` vs `types`? (see https://www.typescriptlang.org/docs/handbook/compiler-options.html) Or they completly different and lib doesn't include typings at all?

mhegazy:
They are completely different. though I could see us moving library components to npm packages in the future, thus eliminating the need to use lib

gertcuykens:
Ok thanks, can you also mention this a bit in the documentation when to use lib?

mhegazy:
It is not deprecated. i am just saying we could publish the built in lib components to npm as well.

pages/tsconfig.json.md
mhegazy: sure

DanielRosenwasser:
:ship: :it:

PR #410, opened 2016-10-08
Fix descriptions of --skip*LibCheck options.

mattmccutchen:


DanielRosenwasser:
Thanks @mattmccutchen!

PR #414, opened 2016-10-12
Mention module concatenation possibilities with TS1.8

ksrimmy:
In the section Namespaces and Modules you mention in the last sentence that it is not possible to concatenate modules to a single JS-File. However with [TypeScript 1.8](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#concatenate-amd-and-system-modules-with---outfile) it is possible to use the compiler switch `--outFile` together with `--module amd` or `--module system`.

### DIFF
--- a/pages/Namespaces and Modules.md
+++ b/pages/Namespaces and Modules.md
@@ -114,4 +114,5 @@ Here's a revised example:
 ## Trade-offs of Modules
 
 Just as there is a one-to-one correspondence between JS files and modules, TypeScript has a one-to-one correspondence between module source files and their emitted JS files.
-One effect of this is that it's not possible to use the `--outFile` compiler switch to concatenate multiple module source files into a single JavaScript file.
+One effect of this is that it's not possible to concatenate multiple module source files depending on the module system you target.
+For instance, you can't use the `outFile` option while targeting `commonjs` or `umd`, but with TypeScript 1.8 and later, [it's possible](https://www.typescriptlang.org/docs/release-notes/typescript-1.8.html#concatenate-amd-and-system-modules-with---outfile) to use `outFile` when targeting `amd` or `system`.


pages/Namespaces and Modules.md
```diff
@@ -114,4 +114,5 @@ Here's a revised example:
 ## Trade-offs of Modules
 
 Just as there is a one-to-one correspondence between JS files and modules, TypeScript has a one-to-one correspondence between module source files and their emitted JS files.
-One effect of this is that it's not possible to use the `--outFile` compiler switch to concatenate multiple module source files into a single JavaScript file.
+One effect of this is that it's not possible to concatenate multiple module source files with every module loader. 
+However, with [TypeScript 1.8](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#concatenate-amd-and-system-modules-with---outfile) it is possible to use the compiler switch `--outFile` together with `--module amd` or `--module system`.
```
DanielRosenwasser: This is good, but I think linking to the site itself might be better. I'd also suggest something like this:

```md
One effect of this is that it's not possible to concatenate multiple module source files depending on the module system you target.
For instance, you can't use the `outFile` option while targeting `commonjs` or `umd`, but with TypeScript 1.8 and later, [it's possible](https://www.typescriptlang.org/docs/release-notes/typescript-1.8.html#concatenate-amd-and-system-modules-with---outfile) to use `outFile` when targeting `amd` or `system`.

DanielRosenwasser:
Looks like the lint failed due to trailing spaces. Would you mind fixing that up?

ksrimmy:
lint works now. I changed the wording to you version.
To what site would you like to link? To the Compiler Options page? There are no anchors in that document... let's see, maybe I create another merge request ;-)

DanielRosenwasser:
@ksrimmy looks good - one thing I'll point out is that the email address your commits used aren't associated with your GitHub account. If it doesn't matter, I can pull this in, but you might you might want to rewrite the commit with a git rebase, using `git commit --amend --author "..."` along the way. Alternatively, you can associate the email address for those commits with GitHub.

Let me know which you'd prefer.

ksrimmy:
Okay, I changed the message and used `git push --force`.
Is there a better option beside `git push --force` to change (hide) an accidentally pushed email address?

DanielRosenwasser:
Well you can create a new branch and GitHub now allows you to change the branch for a given PR, but `--force` has always been my weapon of choice. 😄 

Thanks for the PR!

PR #416, opened 2016-10-16
Fixed broken link in React Webpack tutorial

nilpath:
Updated the path reference to the page about tsconfig.json as it was broken.


DanielRosenwasser:
Thanks! This should be fixed the next time we push out the change.

PR #418, opened 2016-10-17
Fixes typo in "Migrating from JavaScript" md

hegdeashwin:
Fixes #


pages/tutorials/Migrating from JavaScript.md
```diff
@@ -232,7 +232,7 @@ module.exports.feedPets = function(pets) {
 you could write that out as the following:
 
 ```ts
-export function feedPets(pets) {
+exports function feedPets(pets) {
```
mhegazy: this is not valid TS syntax. please see http://www.typescriptlang.org/docs/handbook/modules.html#export for more details.

PR #425, opened 2016-10-26
page(Advanced Types): update code sample

NicBright:
Fixes #

Type assertion is not necessary here. It only increases noise and makes the example more difficult to read.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -13,11 +13,11 @@ Here's a simple example that shows how to create a mixin:
 function extend<T, U>(first: T, second: U): T & U {
     let result = <T & U>{};
     for (let id in first) {
-        (<any>result)[id] = (<any>first)[id];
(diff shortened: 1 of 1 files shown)


mhegazy:
With this change the sample will throw errors under `--noImplicitAny` flag. it is a goal to make sure all samples compile with no errors under `--noImplicitAny`.

NicBright:
> it is a goal to make sure all samples compile with no errors under --noImplicitAny

Ok. But why do you have this goal? Shouldn't the goal be to make the best possible documentation? Increasing strictness on the examples doesn't help for that goal.

mhegazy:
it is confusing when ppl copy a sample and it does not work for them. so we are trying to avoid that too. 

the type assertions are part of the language. so i am not sure i see the two objectives contradicting.

RyanCavanaugh:

PR #436, opened 2016-11-18
Invoke origin setter in Decorators.md

arusakov:


pages/Decorators.md
```diff
@@ -447,7 +447,6 @@ class Line {
 }
 
 function validate<T>(target: any, propertyKey: string, descriptor: TypedPropertyDescriptor<T>) {
-    let set = descriptor.set;
```
mhegazy: you should call the original `set` instead. i believe this was the intention originally.

arusakov:
@mhegazy 
You're right. Updated.

mhegazy:
though i would have made it `set.apply(this, arguments);` but i guess this way it is clearer.

PR #440, opened 2016-11-24
Add TypeScript 2.1 release notes

mhegazy:
//CC @DanielRosenwasser, @sandersn 

Some parts copied verbatim from @DanielRosenwasser's blog.

This includes the content already in https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript.

### DIFF
--- a/pages/release notes/TypeScript 2.1.md
+++ b/pages/release notes/TypeScript 2.1.md
@@ -2,7 +2,7 @@
 
 In JavaScript it is fairly common to have APIs that expect property names as parameters, but so far it hasn't been possible to express the type relationships that occur in those APIs.
 
-Enter Index Type Query or `keysof`;
+Enter Index Type Query or `keyof`;
 An indexed type query `keyof T` yields the type of permitted property names for `T`.
 A `keyof T` type is considered a subtype of `string`.
 
@@ -67,7 +67,7 @@ interface Person {
 }
 ```
 
-A partial verion of it can would be:
+A partial version of it would be:
 
 ```ts
 interface PartialPerson {
@@ -90,7 +90,7 @@ type PartialPerson = Partial<Person>;
 Mapped types are produced by taking a union of literal types, and computing a set of properties for a new object type.
 They're like [list comprehensions in Python](https://docs.python.org/2/tutorial/datastructures.html#nested-list-comprehensions), but instead of producing new elements in a list, they produce new properties in a type.
 
-In addition to `Partial`, Mapped Types can enable expressing many usueful transfomrations on types:
+In addition to `Partial`, Mapped Types can express many useful transformations on types:
 
 ```ts
 // Keep types the same, but make each property to be read-only.
@@ -223,7 +223,7 @@ Compiling and running the output should result in the correct behavior on an ES3
 
 TypeScript injects a handful of helper functions such as `__extends` for inheritance, `__assign` for spread operator in object literals and JSX elements, and `__awaiter` for async functions.
 
-Previously there were two options either:
+Previously there were two options:
 
  1. inject helpers in *every* file that needs them, or
  2. no helpers at all with `--noEmitHelpers`.
--- a/pages/release notes/TypeScript 2.1.md
+++ b/pages/release notes/TypeScript 2.1.md
@@ -1,0 +1,572 @@
+# `keyof` and Lookup Types
+
+Many libraries take advantage of the fact that objects are (for the most part) just a map of strings to values.
+Given what TypeScript knows about each value's properties, there's a set of known strings (or keys) that you can use for lookups.
+
+That's where the `keyof` operator comes in.
+
+```ts
+interface Person {
+    name: string;
+    age: number;
+    location: string;
+}
+
+let propName: keyof Person;
+```
+
+The above is equivalent to having written out
+
+```ts
+let propName: "name" | "age" | "location";
+```
+
+This `keyof` operator is actually called an *index type query*.
+It's like a query for keys on object types, the same way that `typeof` can be used as a query for types on values.
+
+The dual of this is *indexed access types*, also called *lookup types*.
+Syntactically, they look exactly like an element access, but are written as types:
+
+```ts
+interface Person {
+    name: string;
+    age: number;
+    location: string;
+}
+
+let a: Person["age"];
+```
+
+This is the same as saying that `n` gets the type of the `name` property in `Person`.
+In other words:
+
+```ts
+let a: number;
+```
+
+When indexing with a union of literal types, the operator will look up each property and union the respective types together.
+
+```ts
+// Equivalent to the type 'string | number'
+let nameOrAge: Person["name" | "age"];
+```
+
+This pattern can be used with other parts of the type system to get type-safe lookups, serving users of libraries like Ember.
+
+```ts
+function get<T, K extends keyof T>(obj: T, propertyName: K): T[K] {
+    return obj[propertyName];
+}
+
+let x = { foo: 10, bar: "hello!" };
+
+let foo = get(x, "foo"); // has type 'number'
+let bar = get(x, "bar"); // has type 'string'
+
+let oops = get(x, "wargarbl"); // error!
+```
+
+# Mapped Types
+
+Let's say we have a `Person` type:
+
+```ts
+interface Person {
+    name: string;
+    age: number;
+    location: string;
+}
+```
+
+Much of the time, we want to take an existing type and make each of its properties entirely optional.
+With `Person`, we might write the following:
+
+```ts
+interface PartialPerson {
+    name?: string;
+    age?: number;
+    location?: string;
+}
+```
+
+Notice we had to define a completely new type.
+
+Similarly, we might want to perform a shallow freeze of an object:
+
+```ts
+interface FrozenPerson {
+    readonly name: string;
+    readonly age: number;
+    readonly location: string;
+}
+```
+
+Or we might want to create a related type where all the properties are `Promise`s.
+
+```ts
+interface BooleanifiedPerson {
+    name: boolean;
+    age: boolean;
+    location: boolean;
+}
+```
+
+Notice all this repetition - ideally, much of the same information in each variant of `Person` could have been shared.
+
+Let's take a look at how we could write `BooleanifiedPerson` with a *mapped type*.
+
+```ts
+type BooleanifiedPerson = {
+    [P in "name" | "age" | "location"]: boolean
+};
+```
+
+Mapped types are produced by taking a union of literal types, and compute a set of properties for a new object type.
+They're like [list comprehensions in Python](https://docs.python.org/2/tutorial/datastructures.html#nested-list-comprehensions), but instead of producing new elements in a list, they produce new properties in a type.
+
+In the above example, TypeScript uses each literal type in `"name" | "age" | "location"`, and produces a property of that name (e.g. properties named `name`, `age`, and `location`).
+`P` gets bound to each of those literal types (even though it's not used in this example), and gives the property the type `boolean`.
+
+Right now, this new form doesn't look ideal, but we can use the `keyof` operator to cut down on the typing:
+
+```ts
+type BooleanifiedPerson = {
+    [P in keyof Person]: boolean
+};
+```
+
+And then we can generalize it:
+
+```ts
+type Booleanify<T> = {
+    [P in keyof T]: boolean
+};
+
+type BooleanifiedPerson = Booleanify<Person>;
+```
+
+With mapped types, we no longer have to create new partial or readonly variants of existing types either.
+
+```ts
+// Keep types the same, but make every property optional.
+type Partial<T> = {
+    [P in keyof T]?: T[P];
+};
+
+// Keep types the same, but make each property to be read-only.
+type Readonly<T> = {
+    readonly [P in keyof T]: T[P];
+};
+
+// Same property names, but make the value a promise instead of a concreate one
+type Defered<T> = {
+    readonly [P in keyof T]: Promise<T[P]>;
+};
+
+// Wrap proxies around properties of T
+type Proxify<T> = {
+    [P in keyof T]: { get(): T[P]; set(v: T[P]): void }
+};
+```
+
+So instead of defining a completely new type like `PartialPerson`, we can just write `Partial<Person>`.
+Likewise, instead of repeating ourselves with `FrozenPerson`, we can just write `Readonly<Person>`!
+
+## `Partial`, `Readonly`, `Record`, and `Pick`
+
+`Partial` and `Readonly`, as described earlier, are very useful construct.
+You can use them to describe some common JS routines like:
+
+```ts
+function assign<T>(obj: T, props: Partial<T>): void;
+function freeze<T>(obj: T): Readonly<T>;
+```
+
+For that, they are now included by default in the standard library.
+
+We're also including two other utility types as well: `Record` and `Pick`.
+
+```ts
+// From T pick a set of properties K
+function pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K>;
+
+const nameAndAgeOnly = pick(person, "name", "age");  // { name: string, age: number }
+```
+
+```ts
+// For every properties K of type T, transform it to U
+function mapObject<K extends string | number, T, U>(obj: Record<K, T>, f: (x: T) => U): Record<K, U>
+
+const names = { foo: "hello", bar: "world", baz: "bye" };
+const lengths = mapObject(names, s => s.length);  // { foo: number, bar: number, baz: number }
+```
+
+# Object Spread and Rest
+
+TypeScript 2.1 brings support for [ES2017 Spread and Rest](https://github.com/sebmarkbage/ecmascript-rest-spread).
+
+Similar to array spread, spreading an object can be handy to get a shallow copy:
+
+```ts
+let copy = { ...original };
+```
+
+Similarly, you can merge several different objects so that in the following example, `merged` will have properties from `foo`, `bar`, and `baz`.
+
+```ts
+let merged = { ...foo, ...bar, ...baz };
+```
+
+You can also override existing properties and add new ones:
+
+```ts
+let obj = { x: 1, y: "string" };
+var newObj = {...obj, z: 3, y: 4}; // { x: number, y:number, x: number }
+```
+
+The order of specifying spread operations decides what properties end up in the resulting object;
+properties in later spreads "win out" over previously created properties.
+
+Object rests are the dual of object spreads, in that they can extract any extra properties that don't get picked up when destructuring an element:
+
+```ts
+let obj = { x: 1, y: 1, z: 1 };
+let { z, ...obj1 } = obj;
+obj1; // {x: number; y:number};
+```
+
+# Improved `any` Inference
+
+Previously, if TypeScript can't figure out the type of a variable, it will choose the `any` type.
+
+```ts
+let x;      // implicitly 'any'
+let y = []; // implicitly 'any[]'
+
+let z: any; // explicitly 'any'.
+```
+
+With TypeScript 2.1, instead of just choosing `any`, TypeScript will infer types based on what you end up assigning later on.
+
+This is only enabled if `--noImplicitAny` is set.
+
+##### Example
+
+```ts
+let x;
+
+// You can still assign anything you want to 'x'.
+x = () => 42;
+
+// After that last assignment, TypeScript 2.1 knows that 'x' has type '() => number'.
+let y = x();
+
+// Thanks to that, it will now tell you that you can't add a number to a function!
+console.log(x + y);
+//          ~~~~~
+// Error! Operator '+' cannot be applied to types '() => number' and 'number'.
+
+// TypeScript still allows you to assign anything you want to 'x'.
+x = "Hello world!";
+
+// But now it also knows that 'x' is a 'string'!
+x.toLowerCase();
+```
+
+The same sort of tracking is now also done for empty arrays.
+
+A variable declared with no type annotation and an initial value of `[]` is considered an implicit `any[]` variable.
+Each following `x.push(value)`, `x.unshift(value)` or `x[n] = value` operation _evolves_ the type of the variable in accordance with what elements are added to it.
+
+``` ts
+function f1() {
+    let x = [];
+    x.push(5);
+    x[1] = "hello";
+    x.unshift(true);
+    return x;  // (string | number | boolean)[]
+}
+
+function f2() {
+    let x = null;
+    if (cond()) {
+        x = [];
+        while (cond()) {
+            x.push("hello");
+        }
+    }
+    return x;  // string[] | null
+}
+```
+
+## Implicit any errors
+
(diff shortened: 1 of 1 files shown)


pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+function freeze<T>(obj: T): Readonly<T>;
+```
+
+For that, they are now included by default in the standard library.
+
+We're also including two other utility types as well: `Record` and `Pick`.
+
+```ts
+// From T pick a set of properties K
+function pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K>;
```
DanielRosenwasser: declare function

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+dramaticWelcome();
+```
+
+Compiling and running the output should result in the correct behavior on an ES3/ES5 engine.
+
+# Support for external helpers library (`tslib`)
+
+TypeScript injects a handful of helper functions such as `__extends` for inheritance, `__assign` for spread operator in JSX, and `__awaiter` for async functions.
+Previously there were two options either 1. inject helpers in *every* file that needs them or 2. no helpers at all with `--noEmitHelpers`.
```
DanielRosenwasser: Turn these into an ordered list

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+}
+
+dramaticWelcome();
+```
+
+Compiling and running the output should result in the correct behavior on an ES3/ES5 engine.
+
+# Support for external helpers library (`tslib`)
+
+TypeScript injects a handful of helper functions such as `__extends` for inheritance, `__assign` for spread operator in JSX, and `__awaiter` for async functions.
```
DanielRosenwasser: Assign is also used beyond JSX

pages/release notes/TypeScript 2.1.md
DanielRosenwasser: Mention that this was a painpoint if users wanted to keep their bundle size low.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+Previously there were two options either 1. inject helpers in *every* file that needs them or 2. no helpers at all with `--noEmitHelpers`.
+
+TypeScript 2.1 allows for including these files in your project once in a separate module, and the compiler will emit imports to them as needed.
+
+First, install the [`tslib`](https://github.com/Microsoft/tslib):
```
DanielRosenwasser: the tslib utility library

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+npm install tslib
+```
+
+Second, compile your files using `--importHelpers`:
+
+```sh
+tsc --module commonjs --importHelpers a.ts
+```
+
+##### Example
```
DanielRosenwasser: You don't need this heading.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+For input using the object rest `__assign` helper:
```
DanielRosenwasser: So given the following input, the resulting `.js` file will include an import to `tslib` and use the `__assign` helper from it instead of inlining it.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+# Untyped imports
+
+TypeScript has traditionally been overly strict about how you can import modules.
+This was to avoid typos and prevent users from using modules incorrectly.
+
+However, a lot of the time, you might just want to import an existing module that may not have its own `.d.ts` file.
+Previously this was an error.
+Starting with TypeScript 2.1 this should get much easier.
```
DanielRosenwasser: is now much easier.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+##### Example
+
+```ts
+// Succeeds if `node_modules/asdf/index.js` exists
+import { x } from "asdf";
+```
+
+# Improved `any` Inference
+
+Previously, if TypeScript can't figure out the type of a variable, it will choose the `any` type.
```
DanielRosenwasser: Previously, if TypeScript couldn't figure out the type of a variable, it would choose the `any` type.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+x = "Hello world!";
+
+// But now it also knows that 'x' is a 'string'!
+x.toLowerCase();
+```
+
+The same sort of tracking is now also done for empty arrays.
+
+A variable declared with no type annotation and an initial value of `[]` is considered an implicit `any[]` variable.
+Each following `x.push(value)`, `x.unshift(value)` or `x[n] = value` operation _evolves_ the type of the variable in accordance with what elements are added to it.
```
DanielRosenwasser: Use asterisks, not underscores

DanielRosenwasser: "However, each subequent..."

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+    function g() {
+        x;    // Error: Variable 'x' implicitly has an 'any[]' type.
+    }
+}
+```
+
+# Better inference for literal types
+
+String, numeric and boolean literal types (e.g. `"abc"`, `1`, and `true`) were previously inferred only in the presence of an explicit type annotation.
+Starting with TypeScript 2.1, literal types are *always* inferred by default.
```
DanielRosenwasser: This is super-duper misleading. They're inferred by default, but discarded in at least 50% of cases, so this is confusing.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+The type inferred for a `const` variable or `readonly` property without a type annotation is the type of the literal initializer.
+The type inferred for a `let` variable, `var` variable, parameter, or non-`readonly` property with an initializer and no type annotation is the widened literal type of the initializer.
+Where the widened type for a string literal type is `string`, `number` for numeric literal types, `boolean` for `true` or `false` and the containing enum for enum literal types.
+
+##### Example
+
+```ts
+
```
DanielRosenwasser: Remove line

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+const c1 = 1;  // Type 1
```
DanielRosenwasser: Surround each type with a single quote

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+const c1 = "hello";  // Widening type "hello"
+let v1 = c1;  // Type string
+
+const c2: "hello" = "hello";  // Type "hello"
+let v2 = c2;  // Type "hello"
+```
+
+# Use returned values from super calls as 'this'
+
+In ES2015, constructors which return a value (which is an object) implicitly substitute the value of `this` for any callers of `super()`.
```
DanielRosenwasser: "constructors which return an object" might be shorter

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+# Configuration inheritance
+
+Often a project has multiple output targets, e.g. `ES5` and `ES2015`, debug and production or `CommonJS` and `System`;
+Just a few configuration options change between these two targets, and maintaining multiple `tsconfig.json` files can be a hassle.
+
+TypeScript 2.1 supports inheriting configuration using `extends`, where:
+
+* `extends` is a new top-level property in `tsconfig.json` (alongside `compilerOptions`, `files`, `include`, and `exclude`).
+* `extends`' value is a string containing a path to another configuration file to inherit from.
```
DanielRosenwasser: The value of `extends` must be a string...

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+* The configuration from the base file are loaded first, then overridden by those in the inheriting config file.
+* If a circularity is encountered, we report an error.
```
DanielRosenwasser: circularity between configuration files is encountered

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+Invoking the compiler with `--alwaysStrict` causes:
+
+1. Parses all the code in strict mode.
+2. Writes `"use strict";` directive atop every generated file.
+
+Modules are parsed automatically in strict mode.
+The new flag is recommended for non-module code.
+
+# Support for `--target ES2016`, `--target ES2017` and `--target ESNext`
```
DanielRosenwasser: Move this higher up - probably before implicit any inference.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+interface Person {
+    name: string;
+    age: number;
+    location: string;
+}
+
+let propName: keyof Person;
+```
+
+The above is equivalent to having written out
```
sandersn: `keyof Person` is equivalent to ...

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+```ts
+let propName: "name" | "age" | "location";
+```
+
+This `keyof` operator is actually called an *index type query*.
```
sandersn: delete 'actually'

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+When indexing with a union of literal types, the operator will look up each property and union the respective types together.
+
+```ts
+// Equivalent to the type 'string | number'
+let nameOrAge: Person["name" | "age"];
+```
+
+This pattern can be used with other parts of the type system to get type-safe lookups, serving users of libraries like Ember.
```
sandersn: The last part is pretty formal. I'd rewrite to: ", which is great if you use Ember or something like it."

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+```ts
+interface FrozenPerson {
+    readonly name: string;
+    readonly age: number;
+    readonly location: string;
+}
+```
+
+Or we might want to create a related type where all the properties are `Promise`s.
```
sandersn: The example is wrong. It should be Promise, or else the sentence should talk about booleans.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+Let's take a look at how we could write `BooleanifiedPerson` with a *mapped type*.
+
+```ts
+type BooleanifiedPerson = {
+    [P in "name" | "age" | "location"]: boolean
+};
+```
+
+Mapped types are produced by taking a union of literal types, and compute a set of properties for a new object type.
```
sandersn: computing

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+        }
+    }
+    return x;  // string[] | null
+}
+```
+
+## Implicit any errors
+
+One great benefit of this is that you'll see *way fewer* implicit `any` errors when running with `--noImplicitAny`.
+Implicit `any` errors are only reported when the compiler is unable to know the type of a available without a type annotation.
```
sandersn: typo: available &rarr; variable

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+let foo = get(x, "foo"); // has type 'number'
+let bar = get(x, "bar"); // has type 'string'
+
+let oops = get(x, "wargarbl"); // error!
+```
+
+# Mapped Types
+
+Let's say we have a `Person` type:
```
sandersn: For release notes, I think this intro is pretty long because it takes so much time to explain the problem it solves. I think maybe we should switch this discussion with the one in my PR. The handbook is a better place for a long explanation.

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+};
+```
+
+Mapped types are produced by taking a union of literal types, and compute a set of properties for a new object type.
+They're like [list comprehensions in Python](https://docs.python.org/2/tutorial/datastructures.html#nested-list-comprehensions), but instead of producing new elements in a list, they produce new properties in a type.
+
+In the above example, TypeScript uses each literal type in `"name" | "age" | "location"`, and produces a property of that name (e.g. properties named `name`, `age`, and `location`).
+`P` gets bound to each of those literal types (even though it's not used in this example), and gives the property the type `boolean`.
+
+Right now, this new form doesn't look ideal, but we can use the `keyof` operator to cut down on the typing:
```
sandersn: I would just say "Now we can use the `keyof` operator to cut down on the manually specified property names."

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+    [P in keyof T]: { get(): T[P]; set(v: T[P]): void }
+};
+```
+
+So instead of defining a completely new type like `PartialPerson`, we can just write `Partial<Person>`.
+Likewise, instead of repeating ourselves with `FrozenPerson`, we can just write `Readonly<Person>`!
+
+## `Partial`, `Readonly`, `Record`, and `Pick`
+
+`Partial` and `Readonly`, as described earlier, are very useful construct.
```
sandersn: *constructs

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+You can use them to describe some common JS routines like:
+
+```ts
+function assign<T>(obj: T, props: Partial<T>): void;
+function freeze<T>(obj: T): Readonly<T>;
+```
+
+For that, they are now included by default in the standard library.
```
sandersn: For &rarr; because of

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+    [P in keyof T]?: T[P];
+};
+
+// Keep types the same, but make each property to be read-only.
+type Readonly<T> = {
+    readonly [P in keyof T]: T[P];
+};
+
+// Same property names, but make the value a promise instead of a concreate one
+type Defered<T> = {
```
sandersn: isn't Deferred the correct spelling? I'm not sure anymore

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+type Partial<T> = {
+    [P in keyof T]?: T[P];
+};
+
+// Keep types the same, but make each property to be read-only.
+type Readonly<T> = {
+    readonly [P in keyof T]: T[P];
+};
+
+// Same property names, but make the value a promise instead of a concreate one
```
sandersn: typo:concrete

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+
+TypeScript 2.1 brings support for [ES2017 Spread and Rest](https://github.com/sebmarkbage/ecmascript-rest-spread).
+
+Similar to array spread, spreading an object can be handy to get a shallow copy:
+
+```ts
+let copy = { ...original };
+```
+
+Similarly, you can merge several different objects so that in the following example, `merged` will have properties from `foo`, `bar`, and `baz`.
```
sandersn: I would break this into two sentences: "Similarly, you can merge several different objects. In the following example, ..."

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+You can also override existing properties and add new ones:
+
+```ts
+let obj = { x: 1, y: "string" };
+var newObj = {...obj, z: 3, y: 4}; // { x: number, y:number, x: number }
+```
+
+The order of specifying spread operations decides what properties end up in the resulting object;
```
sandersn: decides &rarr; determines

pages/release notes/TypeScript 2.1.md
mhegazy: done

pages/release notes/TypeScript 2.1.md
mhegazy: done

pages/release notes/TypeScript 2.1.md
mhegazy: done and done.

pages/release notes/TypeScript 2.1.md
mhegazy: done

pages/release notes/TypeScript 2.1.md
mhegazy: removed

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done

pages/release notes/TypeScript 2.1.md
mhegazy: done and done.

pages/release notes/TypeScript 2.1.md
mhegazy: added `for `const` variables and `readonly` properties.`

pages/release notes/TypeScript 2.1.md
mhegazy: that is confusing.. it is not type `'1'` it is type `1`

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: Fixed.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: changed

pages/release notes/TypeScript 2.1.md
mhegazy: done.

pages/release notes/TypeScript 2.1.md
mhegazy: I agree. cut down significantly.

mhegazy:
@DanielRosenwasser and @sandersn, I have simplified the notes for `keyof` and mapped type, less chatty now.

pages/release notes/TypeScript 2.1.md
```diff
@@ -1,9 +1,12 @@
 # `keyof` and Lookup Types
 
-Many libraries take advantage of the fact that objects are (for the most part) just a map of strings to values.
-Given what TypeScript knows about each value's properties, there's a set of known strings (or keys) that you can use for lookups.
+In JavaScript it is fairly common to have APIs that expect property names as parameters, but so far it hasn't been possible to express the type relationships that occur in those APIs.
 
-That's where the `keyof` operator comes in.
+Enter Index Type Query or `keysof`;
```
sandersn: typo: `keyof`

pages/release notes/TypeScript 2.1.md
```diff
@@ -78,8 +67,7 @@ interface Person {
 }
 ```
 
-Much of the time, we want to take an existing type and make each of its properties entirely optional.
-With `Person`, we might write the following:
+A partial verion of it can would be:
```
sandersn: delete 'can'

pages/release notes/TypeScript 2.1.md
```diff
@@ -89,78 +77,30 @@ interface PartialPerson {
...
-    [P in keyof T]: boolean
-};
-
-type BooleanifiedPerson = Booleanify<Person>;
-```
+Mapped types are produced by taking a union of literal types, and computing a set of properties for a new object type.
+They're like [list comprehensions in Python](https://docs.python.org/2/tutorial/datastructures.html#nested-list-comprehensions), but instead of producing new elements in a list, they produce new properties in a type.
 
-With mapped types, we no longer have to create new partial or readonly variants of existing types either.
+In addition to `Partial`, Mapped Types can enable expressing many usueful transfomrations on types:
```
sandersn: 1. "enable expressing" &rarr; express
2. typo: useful
3. typo: transformations

pages/release notes/TypeScript 2.1.md
```diff
@@ -283,12 +221,20 @@ Compiling and running the output should result in the correct behavior on an ES3
 
 # Support for external helpers library (`tslib`)
 
-TypeScript injects a handful of helper functions such as `__extends` for inheritance, `__assign` for spread operator in JSX, and `__awaiter` for async functions.
-Previously there were two options either 1. inject helpers in *every* file that needs them or 2. no helpers at all with `--noEmitHelpers`.
+TypeScript injects a handful of helper functions such as `__extends` for inheritance, `__assign` for spread operator in object literals and JSX elements, and `__awaiter` for async functions.
+
+Previously there were two options either:
```
sandersn: I would just delete 'either'. Or say "Previously there were two options; either:"

pages/release notes/TypeScript 2.1.md
```diff
@@ -0,0 +1,513 @@
...
+
+```ts
+interface Person {
+    name: string;
+    age: number;
+    location: string;
+}
+```
+
+A partial verion of it can would be:
```
sandersn: version

mhegazy:
@DanielRosenwasser let me know if you have more comments and i will get them done in a separate PR.

PR #443, opened 2016-11-29
Document index types, indexed access types and mapped types

sandersn:

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -367,9 +367,28 @@ let value: Map<number>['foo']; // number
 
 # Mapped types
 
-Mapped types allow you to create new types based on old types.
-For example, you can make all fields of a type `readonly` or optional.
-Here's how you do it:
+A common task is to take an existing type and make each of its properties optional:
+
+```ts
+interface PersonPartial {
+    name?: string;
+    age?: number;
+}
+```
+
+Or we might want a readonly version:
+
+```ts
+interface PersonReadonly {
+    readonly name: string;
+    readonly age: number;
+}
+```
+
+This happens often enough in Javascript that TypeScript provides a way to create new types based on old types &mdash; mapped types.
+In a mapped type, the new type transforms each property in the old type in the same way.
+For example, you can make all properties of a type `readonly` or optional.
+Here are a couple of examples:
 
 ```ts
 type Readonly<T> = {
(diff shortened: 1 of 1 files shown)


sandersn:
@DanielRosenwasser or @mhegazy mind taking a look?

pages/Advanced Types.md
```diff
@@ -283,6 +283,197 @@ The right side of the `instanceof` needs to be a constructor function, and TypeS
 
 in that order.
 
+# Index types
```
mhegazy: Not sure where is a good place to surface this terminology:

  - `keyof T` => Index Type Query
  - `T[K]` => Indexed Access Type
  - `{ [P in K] : T }` => Mapped Types
    -  ` { [P in keyof T]: F<T[K]>; }`  => Homomorphic Mapped Types

See https://github.com/Microsoft/TypeScript/issues/11943

pages/Advanced Types.md
```diff
@@ -283,6 +283,197 @@ The right side of the `instanceof` needs to be a constructor function, and TypeS
...
+    age: number;
+}
+let person: Person;
+let strings: string[] = pluck(person, ['name']); // ok, string[]
+```
+
+The compiler checks that `name` is actually a property on `Person`, and it knows that `strings` is a `string[]` because `name` is a `string`.
+To make this work, the example introduces a couple of new type operators.
+First is `keyof T`, the index type query operator.
+For any type `T`, `keyof T` is the union of property names of `T`.
```
mhegazy: is the union of **known public** property names of `T`.

pages/Advanced Types.md
```diff
@@ -283,6 +283,197 @@ The right side of the `instanceof` needs to be a constructor function, and TypeS
...
+The syntax is similar to `typeof` except that it works on types instead of values.
```
mhegazy: not sure this line adds much value.

pages/Advanced Types.md
```diff
@@ -283,6 +283,197 @@ The right side of the `instanceof` needs to be a constructor function, and TypeS
...
+
+The second operator is `T[K]`, the indexed access operator.
+Here, the type syntax reflects the expression syntax.
+That means that `person['name']` has the type `Person['name']` &mdash; which in our example is just `string`.
+However, just like index type queries, you can use `T[K]` in a generic context, which is where its real power comes to life.
+You just have to make sure that the type variable `K extends keyof T`.
+Here's another example with a function named `getProperty`.
+
+```ts
+function getProperty<T, K extends keyof T>(o: T, name: K): K[T] {
```
mhegazy: typo: return type should be `T[K]`

pages/Advanced Types.md
```diff
@@ -283,6 +283,197 @@ The right side of the `instanceof` needs to be a constructor function, and TypeS
...
+Or we might want a readonly version:
+
+```ts
+interface PersonReadonly {
+    readonly name: string;
+    readonly age: number;
+}
+```
+
+This happens often enough in Javascript that TypeScript provides a way to create new types based on old types &mdash; mapped types.
```
mhegazy: why `&mdash;` and not just `-`?

pages/Advanced Types.md
```diff
@@ -283,6 +283,197 @@ The right side of the `instanceof` needs to be a constructor function, and TypeS
...
+
+```ts
+type PersonPartial = Partial<Person>;
+type ReadonlyPerson = Readonly<Person>;
+```
+
+Let's take a look at the simplest mapped type and its parts:
+
+```ts
+type Properties = 'option1' | 'option2';
```
mhegazy: i would call this `Keys`.

pages/Advanced Types.md
```diff
@@ -283,6 +283,197 @@ The right side of the `instanceof` needs to be a constructor function, and TypeS
...
+
+```ts
+type Pick<T, K extends keyof T> = {
+    [P in K]: T[P];
+}
+type Record<K extends string | number, T> = {
+    [P in K]: T;
+}
+```
+
```
mhegazy: we should add some notes about inferring from mapped types. a common question is how to get the "un-proxified" or "un-partial" version of a value. see https://github.com/Microsoft/TypeScript/issues/12578#issuecomment-263967420 for an example.

Also worth noting how inference works from homomorphoic types, see https://github.com/Microsoft/TypeScript/pull/12528

pages/Advanced Types.md
sandersn: I mention the names below in the text (except homomorphic mapped types) but they're kind of buried. I'll think about ways to call them out.

pages/Advanced Types.md
sandersn: fixed

pages/Advanced Types.md
sandersn: true

pages/Advanced Types.md
sandersn: `-` isn't long enough and `--` doesn't look good when rendered to HTML.

pages/Advanced Types.md
sandersn: sure

pages/Advanced Types.md
sandersn: good idea. I will add that shortly.

sandersn: Done

PR #446, opened 2016-12-08
Does not require `this.` before props in the stateless example

d4nyll:
If it remains `this.props.xxx`, it will throw `Uncaught TypeError: Cannot read property 'compiler' of undefined` on `Hello.tsx:5`


DanielRosenwasser:
Thanks!

PR #449, opened 2016-12-08
fixed typos in code

grzegorz-zur:


mhegazy:
The error seems intentional to me. @DanielRosenwasser  thoughts?

DanielRosenwasser:
Yes, that's the goal - sorry if you missed the intent @grzegorz-zur .

PR #451, opened 2016-12-08
Mention there are currently 2 webpack loaders for typescript and link…

### LINKED ISSUE #450: Consistently use the same typescript loader for webpack and mention alternatives
Currently in the React & Webpack tutorial the docs specify using the awesome-typescript-loader, Migrating from javascript specifies using the ts-loader.

The docs should be consistent about which loader to use while also provide some additional information about the alternative so the end user could decide which loader is best for him.

morsdyce:
Fixes #450

Consistently use awesome-webpack-loader across documentation and mention there is an alternative.
As well as include a link to more information about the differences between the 2 loaders.

### DIFF
--- a/pages/tutorials/Migrating from JavaScript.md
+++ b/pages/tutorials/Migrating from JavaScript.md
@@ -123,17 +123,12 @@ module.exports = {
 };
 ```
 
-It's important to note several things:
+It's important to note that awesome-typescript-loader will need to run before any other loader that deals with `.js` files.
 
-`awesome-typescript-loader` will need to run before any other loader that deals with `.js` files.
+The same goes for [ts-loader](https://github.com/TypeStrong/ts-loader), another TypeScript loader for Webpack.
+You can read more about the differences between the two [here](https://github.com/s-panferov/awesome-typescript-loader#differences-between-ts-loader).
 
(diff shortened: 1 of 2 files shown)


pages/tutorials/Migrating from JavaScript.md
```diff
@@ -123,9 +123,18 @@ module.exports = {
 };
 ```
 
-It's important to note that `ts-loader` will need to run before any other loader that deals with `.js` files.
+It's important to note several things:
```
DanielRosenwasser: Could you change these lines to the following?

```md
It's important to note that awesome-typescript-loader will need to run before any other loader that deals with `.js` files.

The same goes for [ts-loader](https://github.com/TypeStrong/ts-loader), another TypeScript loader for Webpack.
You can read more about the differences between the two [here](https://github.com/s-panferov/awesome-typescript-loader#differences-between-ts-loader).

You can see an example of using Webpack in our [tutorial on React and Webpack](./React & Webpack.md).
```

morsdyce:
Updated, thanks for helping with the phrasing, need to work on my writing skills. 👍

DanielRosenwasser:
Thanks a bunch for the PR!

PR #456, opened 2016-12-11
Fix two small errors in React + Webpack tutorial.

tonycosentini:
This fixes two issues I ran into while reading through the tutorial.

The first issue is that `node_modules` needs to be explicitly excluded, otherwise you end up with a bunch of "Cannot redeclare..." errors. (I'm not sure if this is the best solution here, feel free to let me know if there is a better way to handle this.)

The second issue is that the Hello component wasn't exported, resulting in a compile error.

### DIFF
--- a/pages/tutorials/React & Webpack.md
+++ b/pages/tutorials/React & Webpack.md
@@ -101,6 +101,9 @@ Simply create a new file in your project root named `tsconfig.json` and fill it 
     },
     "include": [
         "./**/*"
+    ],
+    "exclude": [
+        "node_modules"
     ]
 }
 ```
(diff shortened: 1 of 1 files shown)


mhegazy:
I do not see how the explicit exclude helps.. the default exclude has node_modules as well as the outdir. Can u share more context?

hesselink:
I just had the same issue. The "node_modules" directory doesn't seem to be excluded by default, resulting in a huge list of errors. The [docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) do mention that "node_modules" should be excluded by default, so maybe this is a bug and not a documentation issue?

tonycosentini:
Apologies for never following up here, but thanks for merging :)

 I think the main issue is that everything is included for webpack, not just the app source. Another approach would be to only include the `src` dir, then the `node_modules` exclude wouldn't be required. Either approach works, I'm not sure which one is more clean.

PR #459, opened 2016-12-15
correction

mindplay-dk:
Fixes #

the React identifier is not hard-coded

### DIFF
--- a/pages/JSX.md
+++ b/pages/JSX.md
@@ -25,7 +25,7 @@ Mode       | Input     | Output                       | Output File Extension
 
 You can specify this mode using either the `--jsx` command line flag or the corresponding option in your [tsconfig.json](./tsconfig.json.md) file.
 
-> *Note: The identifier `React` is hard-coded, so you must make React available with an uppercase R.*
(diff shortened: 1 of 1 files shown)


pages/JSX.md
```diff
@@ -25,7 +25,7 @@ Mode       | Input     | Output                       | Output File Extension
 
 You can specify this mode using either the `--jsx` command line flag or the corresponding option in your [tsconfig.json](./tsconfig.json.md) file.
 
-> *Note: The identifier `React` is hard-coded, so you must make React available with an uppercase R.*
+> *Note: The react namespace (the object invoked for `createElement` and `__spread` when targeting `"react"` JSX emit) defaults to `React` (with an upper-case "R") - this can be changed using the `--reactNamespace` [compiler option](https://www.typescriptlang.org/docs/handbook/compiler-options.html).*
```
mhegazy: `__spread` is not used any longer.

mhegazy: Since we are touching this, we should also reference `--jsxFactory` that can override the `React.createElement` all together to a different function, e.g. `h`, or `myHelper.createElement`

RyanCavanaugh:
No response

PR #463, opened 2016-12-22
Update tsconfig.json

sethjuarez:
I got a bunch of errors when running this sample. I changed the include attribute as described and everything worked.

Fixes #


mhegazy:
should not be needed any longer. https://github.com/Microsoft/TypeScript-Handbook/commit/a95187c4f4966053b9c7032c94fd0c678b1837bc added the exclude.

sethjuarez:
I see it! Makes sense - thanks.

PR #467, opened 2016-12-28
Don't ask people to add their packages to `notNeededPackages.json` on DefinitelyTyped

andy-ms:
This was added by 04f99c3.
`types-publisher` uses `notNeededPackages.json` to know what `@types` packages to deprecate. If there is no such `@types` package it will have problems.

Sorry for the confusion IdentityModel/oidc-client-js#135!


pages/declaration files/Publishing.md
```diff
@@ -84,13 +84,6 @@ If your type definitions depend on another package:
 * *Don't* copy the declarations in your package either.
 * *Do* depend on the npm type declaration package if it doesn't package its declaration files.
 
-## Publicize your declaration file
-
-After publishing your declaration file with your package, make sure to add a reference to it in the [DefinitelyTyped repo external package list](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/notNeededPackages.json).
-Adding this will allow search tools to know that your package provides its own declarations.
```
DanielRosenwasser: This is why I originally sent a PR - are you certain that we have no plan to do so? @RyanCavanaugh

mhegazy: From our experience so far, most package authors do not add an entry in this file. we should alternatively have a crawling job that looks in the npm registry for these packages for search purposes.

DanielRosenwasser: @mhegazy would that work for scoped packages?

mhegazy: scoped packages are not supported in `@types` anyways.

andy-ms: I added a script to check all NPM packages for types. We don't do anything useful with the output yet, though. Could be used to extend the `create-search-index` and `publish-registry` steps.

andy-ms:
Aparently this was done in https://github.com/Microsoft/TypeScript-Handbook/commit/17098fde05099b0d70cc6be0f0420cc10ea0de38.

PR #469, opened 2016-12-30
Fix mismatch in description of discriminated unions

vgrigoriu:
Probable typo: the text mentions four ingredients of discriminated unions, but only three are enumerated.


DanielRosenwasser:
If only there was a type-checker for English. 😉 

Thanks for the PR!

PR #470, opened 2016-12-31
Fix incorrect language of codeblock in namespaces

azu:
Fix incorrect source language in https://www.typescriptlang.org/docs/handbook/namespaces.html

namespaces typescript 2016-12-31 10-21-08

`ts` => `html`.


mhegazy:
thanks!

PR #471, opened 2016-12-31
Added in some of the great info about non-homomorphic transformations…

JakeGinnivan:
… in type mapping from https://github.com/Microsoft/TypeScript/issues/13224#issuecomment-269807806

Just trying to extend the docs for something I could not figure out how to do after reading the docs on type mapping. // @mhegazy

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -794,7 +794,7 @@ type ReadonlyPerson = Readonly<Person>;
 ```
 
 Property mapping is a [homomorphic](https://en.wikipedia.org/wiki/Homomorphism) transformation, so if you wanted to do the
- opposite of the `Partial` type and make all optional properties required (a non-homomorphic transformation) you could create 
+ opposite of the `Partial` type and make all optional properties required (a non-homomorphic transformation) you could create
  a type `EnsureAll`
 
 ```ts
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -793,6 +793,29 @@ type PersonPartial = Partial<Person>;
 type ReadonlyPerson = Readonly<Person>;
 ```
 
+Property mapping is a [homomorphic](https://en.wikipedia.org/wiki/Homomorphism) transformation, so if you wanted to do the
+ opposite of the `Partial` type and make all optional properties required (a non-homomorphic transformation) you could create 
+ a type `EnsureAll`
+
+```ts
+type EnsureAll<T, K extends string> = {
+    [P in K]: T[P];
+}
+```
+
+And to use it:
+
+```ts
+interface Person {
+    name: string;
+    age?: number;
+}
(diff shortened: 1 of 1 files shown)


mhegazy:
@sandersn can you take a look

pages/Advanced Types.md
```diff
@@ -793,6 +793,29 @@ type PersonPartial = Partial<Person>;
 type ReadonlyPerson = Readonly<Person>;
 ```
 
+Property mapping is a [homomorphic](https://en.wikipedia.org/wiki/Homomorphism) transformation, so if you wanted to do the
```
sandersn: Homomorphic mapped types are already described briefly in the "Inference from mapped types" section, so that section needs to be updated as well.

pages/Advanced Types.md
```diff
@@ -793,6 +793,29 @@ type PersonPartial = Partial<Person>;
...
+```ts
+interface Person {
+    name: string;
+    age?: number;
+}
+type PersonWithAge = EnsureAll<Person, keyof Person> // { name: string, age: number }
+```
+
+Notice the constraint on K is string; this way the type system can not assert that the result of this
+ transformation is a homomorphic mapping on T, and thus no modifiers are copied through.
```
sandersn: Need an example of coping vs not-copying.

sandersn:
I think my requests amount to rewriting the whole PR, so I made a PR at #473 that does more or less that. @JakeGinnivan @mhegazy do you want to take a look?

I like the `EnsureAll` example but I ended up using `Pick` and `Record` as examples since they were already discussed as part of the built-in types.

JakeGinnivan:
Sounds good to me @sandersn, was just an area where I had some difficulty making it work the way I wanted.

pages/Advanced Types.md
```diff
@@ -793,6 +793,29 @@ type PersonPartial = Partial<Person>;
 type ReadonlyPerson = Readonly<Person>;
 ```
 
+Property mapping is a [homomorphic](https://en.wikipedia.org/wiki/Homomorphism) transformation, so if you wanted to do the
+ opposite of the `Partial` type and make all optional properties required (a non-homomorphic transformation) you could create
+ a type `EnsureAll`
+
+```ts
+type EnsureAll<T, K extends string> = {
```
masaeedu: Did you actually test this `EnsureAll` type? I get a type error saying you can't index `T` with `K`, presumably because `K` isn't constrained to the keys of `T`.

masaeedu:
@sandersn Could you please add an example for `EnsureAll` as well? It isn't clear how to do the reverse transformation of `Partial`, and I suspect it might not be possible in the current type system.

I'm testing the following with 2.2.2, and can't get it to work:

```ts
type Required<T, K extends keyof T> = {
    [P in K]: T[P]
};

interface X {
    y?: string 
}

let safeX: Required<X, keyof X> = {}; // Assignment should have failed
let y: string = safeX.y;              // Error: string | undefined is not assignable to string
```

PR #477, opened 2017-01-05
fixed variable declaration in namespaces.md

livioc:
there was a wrong 'var' usage


mhegazy:
It is not wrong really. Var is a valid variable declaration. But let is consistent either the rest of the guide style .

livioc:
Thanks, anyway it's a variable declaration inside a for-loop so, at least in ts 2.1, this is a true error ("Cannot redeclare block-scoped variable 'name'")

PR #478, opened 2017-01-06
Fix hyperlink to 1.8 release notes

styfle:
Fixes dead link 

> The resource you are looking for has been removed, had its name changed, or is temporarily unavailable.

It looks like the handbook got moved and this url was never updated.


DanielRosenwasser:
thanks!

PR #493, opened 2017-01-28
Fix typo in last example in Generics

mightyiam:
I don't know all the syntaxes of TypeScript, but this seems like a typo.

### DIFF
--- a/pages/Generics.md
+++ b/pages/Generics.md
@@ -319,9 +319,7 @@ class Lion extends Animal {
     keeper: ZooKeeper;
 }
 
-function findKeeper<A extends Animal, K> (a: {new(): A;
-    prototype: {keeper: K}}): K {
-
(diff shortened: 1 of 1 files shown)


pages/Generics.md
```diff
@@ -319,9 +319,7 @@ class Lion extends Animal {
     keeper: ZooKeeper;
 }
 
-function findKeeper<A extends Animal, K> (a: {new(): A;
-    prototype: {keeper: K}}): K {
```
mhegazy: not sure i see why `;` is a typo?

mightyiam: Isn't that `;` inside an object literal, where a `,` should be? And the line drop also sees inconsistent.

mightyiam: Or is that not an object literal, but an interface?

mhegazy: both `;` and `,` are allowed in object literal types.

```ts
var y: { a: number; b, c: string };
```

PR #494, opened 2017-01-30
Update MSBuild options on newer compiler option

yuit:
Fix
https://github.com/Microsoft/TypeScript/issues/13706
https://github.com/Microsoft/TypeScript/issues/13653

Should merged after we release 2.2 since the fix will be in that release (unless we put it in NuGet)

### DIFF
--- a/pages/Compiler Options in MSBuild.md
+++ b/pages/Compiler Options in MSBuild.md
@@ -26,27 +26,32 @@ Compiler Option                              | MSBuild Property Name            
 `--allowSyntheticDefaultImports`             | TypeScriptAllowSyntheticDefaultImports     | boolean
 `--allowUnreachableCode`                     | TypeScriptAllowUnreachableCode             | boolean
 `--allowUnusedLabels`                        | TypeScriptAllowUnusedLabels                | boolean
+`--alwaysStrict`                             | TypeScriptAlwaysStrict                     | boolean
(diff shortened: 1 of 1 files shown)


pages/Compiler Options in MSBuild.md
```diff
@@ -26,27 +26,32 @@ Compiler Option                              | MSBuild Property Name
...
 `--experimentalDecorators`                   | TypeScriptExperimentalDecorators           | boolean
 `--forceConsistentCasingInFileNames`         | TypeScriptForceConsistentCasingInFileNames | boolean
 `--help`                                     | *Not supported in MSBuild*                 |
+`--importHelpers`                            | TypeScriptImportHelpers                    | boolean
 `--inlineSourceMap`                          | TypeScriptInlineSourceMap                  | boolean
 `--inlineSources`                            | TypeScriptInlineSources                    | boolean
 `--init`                                     | *Not supported in MSBuild*                 |
 `--isolatedModules`                          | TypeScriptIsolatedModules                  | boolean
 `--jsx`                                      | TypeScriptJSXEmit                          | `React` or `Preserve`
+`--jsxFactory`                               | TypeScriptJSXFactory                       | boolean
```
mhegazy: `string` and not `boolean`. more accurately it should be an qualified name.

PR #498, opened 2017-02-04
Add missing assignment in Index Types example

mightyiam:
`module.ts(10,31): error TS2454: Variable 'person' is used before being assigned.`


mightyiam:
@yarom82

PR #499, opened 2017-02-04
Remove wrong sentence in Index Types section

mightyiam:
@yarom82


mhegazy:
The text is acctually correct. but i see why it can be confusing.

PR #510, opened 2017-02-15
Added webpack 2 example

mauricedb:
Added a Webpack 2 configuration example as the syntax is slightly different with `rules` instead of `loaders`.


pages/Integrating with Build Tools.md
```diff
@@ -167,6 +167,27 @@ module.exports = {
...
+        path: '/',
+        filename: "bundle.js"
+    },
+    resolve: {
+        extensions: [".tsx", ".ts", ".js"]
+    },
+    module: {
+        rules: [
+            // all files with a '.ts' or '.tsx' extension will be handled by 'ts-loader'
+            { test: /\.tsx?$/, loader: "ts-loader", exclude: /node_modules/ }
```
TheLarkInn: Please change loader to use

pages/Integrating with Build Tools.md
```diff
@@ -167,6 +167,27 @@ module.exports = {
...
+
+```js
+module.exports = {
+    entry: "./src/index.tsx",
+    output: {
+        path: '/',
+        filename: "bundle.js"
+    },
+    resolve: {
+        extensions: [".tsx", ".ts", ".js"]
```
TheLarkInn: This is an incorrect "appending" from the default values. See webpack.js.org/configuration/resolve

pages/Integrating with Build Tools.md
DanielRosenwasser: For the record, I think @TheLarkInn is referring to https://webpack.js.org/configuration/module/#rule-use

pages/Integrating with Build Tools.md
```diff
@@ -167,6 +167,27 @@ module.exports = {
 }
 ```
 
+### Basic webpack.config.js when using Webpack 2
```
DanielRosenwasser: Can you swap sections for Webpack 1 & 2?

mauricedb:
@TheLarkInn and @DanielRosenwasser 

I have updated the PR with the requested changes. Please let me know if this is what you were looking for.

Also rebased the PR on master.

Maurice

mauricedb:
@DanielRosenwasser Is this ever going to get merged? The docs still only show the Webpack 1  configuration and we are at version 4 already.

PR #514, opened 2017-02-19
update webpack and react tutorial to webpack v2

aleburato:
Running this tutorial with Webpack v. >2.1 throws an error:

```
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.module has an unknown property 'preLoaders'. These properties are valid:
   object { exprContextCritical?, exprContextRecursive?, exprContextRegExp?, exprContextRequest?, loaders?, noParse?, rules?, unknownContextCritical?, unknownContextRecursive?, unknownContextRegExp?, unknownContextRequest?, unsafeCache?, wrappedContextCritical?, wrappedContextRecursive?, wrappedContextRegExp? }
   Options affecting the normal modules (`NormalModuleFactory`).
 - configuration.resolve.extensions[0] should not be empty.
```

I have updated the tutorial and added the correct code for webpack v2.

### DIFF
--- a/pages/tutorials/React & Webpack.md
+++ b/pages/tutorials/React & Webpack.md
@@ -190,6 +190,49 @@ Facebook makes CDN-hosted versions of React available, and you can [read more ab
 
 Create a `webpack.config.js` file at the root of the project directory.
 
+Note: Webpack v.1.* and v2.* need different configuration, as v2 introduced breaking changes.
+
+## Webpack 2 configuration file
+
+```js
+module.exports = {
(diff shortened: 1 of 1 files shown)


RyanCavanaugh:
Trying to write up how to use Webpack in our docs seems to have put us on a really fast treadmill that we're not actually equipped to keep up with. I'm going to close the old PRs relating to Webpack on the assumption that they're probably just as out-of-date as the content they're trying to update. Sorry to have let this sit around so long.

PR #518, opened 2017-02-23
Set text as note in interface.md

anteriovieira:
The passage described below sounds like a note.

It's worth pointing out that the type-checker does not require that these properties come in any sort of order, only that the properties the interface requires are present and have the required type.


RyanCavanaugh:

PR #521, opened 2017-02-27
Fix object type statement

alexjoverm:
Since TypeScript 2.2, you can use object type for non-primitives, so that sentence wasn't true anymore

This fixes an issue opened in TypeScript repo: https://github.com/Microsoft/TypeScript/issues/14340#issuecomment-282836413


mhegazy:
We will need to add a section on `object` in https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Basic%20Types.md. Will file a new issue for that.

alexjoverm:
Thanks, looks good like that :)

PR #525, opened 2017-03-03
Update Basic Types.md

zsong:
Fixes #


mhegazy:
Template string literals are allowed to have new lines. this one of the reasons the templates were added to JS. The sample demonstration this ability.

PR #528, opened 2017-03-14
add example for class decorator

gertcuykens:
It took me 3 days just to understand this **** :P I believe some decorators need way more attention to explain properly

### DIFF
--- a/pages/Decorators.md
+++ b/pages/Decorators.md
@@ -163,6 +163,28 @@ function sealed(constructor: Function) {
 
 When `@sealed` is executed, it will seal both the constructor and its prototype.
 
+Next we have a example how to override the constructor.
+
+```ts
+function classDecorator<T extends {new(...args:any[]):{}}>(constructor:T) {
+    return class extends constructor {
+        newProperty = "new property";
+        hello = "override";
+    }
+}
+
+@classDecorator
+class Greeter {
+    property = "property";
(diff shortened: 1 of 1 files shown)


pages/Decorators.md
```diff
@@ -163,6 +163,41 @@ function sealed(constructor: Function) {
 
 When `@sealed` is executed, it will seal both the constructor and its prototype.
 
+Next we have a example how to override the constructor.
+
+```ts
```
mhegazy: We should use the [class mixin pattern that is now supported in TS 2.2](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#support-for-mix-in-classes):

```ts
 function classDecorator<T extends {new(...args:any[]): {}}>(constructor:T ) {
     return class extends constructor {
         newProperty = "new property";
         hello = "override";
     }
 }
 
 @classDecorator
 class Greeter {
     property = "property";
     hello: string;
     constructor(m: string) {
         this.hello = m;
     }
 }
 
 console.log(new Greeter("world"));
```

gertcuykens: DOH! XD ok let me test this first :P

gertcuykens: Ok tested it, changed the example thanks :) I must admit way easier to understand what's going on then my convoluted function ordeal

PR #530, opened 2017-03-17
Update info in --module compiler option

### LINKED ISSUE #428: 'Module' option: 'es6' and 'es2015' values may not be used when targeting ES5 or lower
According to https://github.com/Microsoft/TypeScript/issues/6319 it's now allowed to have `--module es2015` and `--target es5`, but page https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Compiler%20Options.md for `module` option stays:
`'es6' and 'es2015' values may not be used when targeting ES5 or lower.`

marzelin:
Fixes #428


marzelin:
In addition it would be helpful to add a suggestion ie: 'also add `moduleResolution: node` because by default it is set to `classic` which is probably not what you want`

PR #535, opened 2017-03-29
Missing and misplaced semicolons

magul:
- As long as we don't use JavaScript Standard Style JavaScript statement should end with semicolon (there is Automatic Semicolon Insertion, but that's fixing issue, not make it irrelevant)
- Enums declaration (like function declaration, class declaration, interface declaration) should not end with semicolon (see e.g.: [Enums.md](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Enums.md) or [TypeScript Language Specification](https://github.com/Microsoft/TypeScript/blob/master/doc/TypeScript%20Language%20Specification.pdf)).

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -43,14 +43,14 @@ let fullName: string = `Bob Bobbington`;
 let age: number = 37;
 let sentence: string = `Hello, my name is ${ fullName }.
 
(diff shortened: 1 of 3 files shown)


mhegazy:
thanks!

PR #537, opened 2017-04-01
Correct a mistake in the original text

lifuhuang:
Every protected/private member in target type must exist in the source instance, not the other way around.

Fixes #

### DIFF
--- a/pages/Type Compatibility.md
+++ b/pages/Type Compatibility.md
@@ -202,7 +202,7 @@ s = a;  //OK
 ## Private and protected members in classes
 
 Private and protected members in a class affect their compatibility.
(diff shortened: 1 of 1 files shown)


pages/Type Compatibility.md
```diff
@@ -202,7 +202,7 @@ s = a;  //OK
 ## Private and protected members in classes
 
 Private and protected members in a class affect their compatibility.
-When an instance of a class is checked for compatibility, if the instance contains a private member, then the target type must also contain a private member that originated from the same class.
+When an instance of a class is checked for compatibility, if the target type contains a private member, then the instance must also contain a private member that originated from the same class.
```
mhegazy: I would use `source` instead of `instance` to avoid confusion, i.e. `then the source must`

PR #538, opened 2017-04-03
Fix a bug in sample code.

lifuhuang:
The original code sample was wrong and would throw run-time error. That's because `keeper`, as an instance property, is not in the prototype of the instance's constructor, so I changed `keeper` to be a getter of `Animal`/`Bee`.

Fixes #

### DIFF
--- a/pages/Generics.md
+++ b/pages/Generics.md
@@ -322,11 +322,9 @@ class Lion extends Animal {
     }
 }
 
-function findKeeper<A extends Animal, K> (a: {new(): A;
-    prototype: {keeper: K}}): K {
-
-    return a.prototype.keeper;
+function findKeeper<A extends Animal, K> (a: A & { keeper: K }): K {
+    return a.keeper;
 }
 
-findKeeper(Lion).nametag;  // typechecks!
+findKeeper(new Lion()).nametag;  // typechecks!
 ```
--- a/pages/Generics.md
+++ b/pages/Generics.md
@@ -309,11 +309,17 @@ class Animal {
 }
 
 class Bee extends Animal {
-    keeper: BeeKeeper;
+    private _keeper = new BeeKeeper();
+    get keeper() {
+        return this._keeper;
+    };
 }
 
 class Lion extends Animal {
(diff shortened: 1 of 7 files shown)


mhegazy:
The `findKeeper` seems like a convoluted way to show generics. i would change the function instead to take an instance instead of the class. i.e. `findKeeper(new Lion())`

lifuhuang:
Sorry for the late response. A new commit has been made.

pages/Generics.md
```diff
@@ -309,18 +309,22 @@ class Animal {
...
+    get keeper() {
+        return this._keeper;
+    }
 }
 
-function findKeeper<A extends Animal, K> (a: {new(): A;
-    prototype: {keeper: K}}): K {
-
-    return a.prototype.keeper;
+function findKeeper<A extends Animal, K> (a: A & { keeper: K }): K {
```
mhegazy: not sure why you need the animal all togather.. 

```ts
function findKeeper<T> (a: { keeper: T }): T
```

pages/Generics.md
```diff
@@ -309,18 +309,22 @@ class Animal {
...
-    keeper: BeeKeeper;
+    private _keeper = new BeeKeeper();
+    get keeper() {
+        return this._keeper;
+    };
 }
 
 class Lion extends Animal {
-    keeper: ZooKeeper;
+    private _keeper = new ZooKeeper();
```
mhegazy: no need for complicating the class by adding a private fields and a getter, just keep the property declaration that was there before: ` keeper: ZooKeeper`.

pages/Generics.md
```diff
@@ -309,18 +309,22 @@ class Animal {
 }
 
 class Bee extends Animal {
-    keeper: BeeKeeper;
+    private _keeper = new BeeKeeper();
```
mhegazy: no need for complicating the class by adding a private fields and a getter, just keep the property declaration that was there before: ` keeper: BeeKeeper`.

mhegazy:
hmm.. looks like something went wrong there.. github is not happy :(

image

PR #539, opened 2017-04-03
Clarified and expanded section on rootDirs

### LINKED ISSUE #483: Clarify capabilities of the rootDirs option
`rootDirs` is a really powerful feature. I love it. It allows some very precise runtime behavior to be modeled but I think this could be more discoverable.
Recently I attempted to model [JSPM's conditional substitution](http://jspm.io/0.17-beta-guide/conditional-substitution.html) feature.

Essentially this involves an import such as
```ts
import salutation from './#{lang}/text';
```
The directory layout I used was
**src/**
--- **main.tsx**
--- **en/**
------- **text.tsx**
---**zh/**
------- **text.tsx**

While I could have used an ambient external module declaration, I figured there would be a way to get TypeScript to resolve one of the files containing the conditional import by substituting either **zh** or **en** statically and resolving that path.
There is a way but it is not at all clear that it will work. I essentially stumbled on it by accident.
```js
{
  "compilerOptions": {
    "module": "system",        
    "target": "es2015",
    "rootDirs": [
      // only one of zh or en is actually needed
      "src/zh",
      "src/en",
      "src/#{lang}"
    ]
  }
}
```
It would be great if the documentation were expanded to explain this as it seems non-obvious that the non-existent directory **src/#{lang}** would even be a valid entry. The configuration above was the last one I tried and I did not expect it to work.

aluanhaddad:
Clarified and expanded section on `rootDirs`.
Added series of relatively minor expansions to clarify aspects of `rootDirs`, especially its
capabilities, limitations, and its interaction with logically related features.
Added a simple example to demonstrate abitrary path merging in relative module
specifers.
Clarified interaction (or lack there of) with paths.
Clarify interation with --moduleResolution.

Fixes #483

### DIFF
--- a/pages/Module Resolution.md
+++ b/pages/Module Resolution.md
@@ -352,12 +352,6 @@ By leveraging `rootDirs` we can inform the compiler of this mapping and thereby 
 
 The compiler will now resolve `import messages from './#{locale}/messages'` to `import messages from './zh/messages'` for tooling purposes, allowing development in a locale agnostic manner without compromising design time support.
 
-#### Interaction with other module resolution options
-
-`rootDirs` configuration only applies to literally relative imports such as `import home from "./src/views/home"`. It has no interaction with `paths` even when the latter feature has been used to map absolute imports to what would otherwise be relative ones. For example `rootDirs` ceases to apply if we changed the previous import to `import home from "views/home"` by mapping `views/*` to `src/views/*` via `paths`.
-
-`rootDirs` _is_ impacted by the `moduleResolution` setting with `node` and `classic` exhibiting their respective behaviors and resolving `./module/index` and `./module/module` as expected.
-
 ## Tracing module resolution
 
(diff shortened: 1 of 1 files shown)


pages/Module Resolution.md
```diff
@@ -319,6 +319,41 @@ So following our example, the `tsconfig.json` file should look like:
...
+      "src/de",
+      "src/#{locale}"
+    ]
+  }
+}
+```
+
+The compiler will now resolve `import messages from './#{locale}/messages'` to `import messages from './zh/messages'` for tooling purposes, allowing development in a locale agnostic manner without compromising design time support.
+
+#### Interaction with other module resolution options
```
mhegazy: I do not think this section adds any new information. other features do not have an "interaction with.." section either.

mhegazy:
thanks!

PR #543, opened 2017-04-05
Updated JSX ReadMe for DefinitelyTyped url for React

ryancormack:
Fix typo: DefinitelyTyped has all its typings under a 'typed' folder at the root of the project. Changed the url to the React typings so it goes to the new, correct, location

Fixes #


mhegazy:
thanks!

PR #545, opened 2017-04-07
Remove extensions that are invalid or unnecessary in Webpack 2 config

### LINKED ISSUE #544: Webpack+Typescript Example Doesn't Work
Hey guys, I noticed your example for Webpack+Typescript just doesn't work. Maybe it's for a previous version of Webpack?

https://www.typescriptlang.org/docs/handbook/integrating-with-build-tools.html#webpack

Using this webpack.config.js will result in the error:

```
PS E:\webpack-test> webpack
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.resolve.extensions[0] should not be empty.
```

Removing the first index of the array ("") will fix it.

jlost:
Fixes #544

### DIFF
--- a/pages/Integrating with Build Tools.md
+++ b/pages/Integrating with Build Tools.md
@@ -156,7 +156,7 @@ module.exports = {
     },
     resolve: {
         // Add '.ts' and '.tsx' as a resolvable extension.
-        extensions: ["", ".webpack.js", ".web.js", ".ts", ".tsx", ".js"]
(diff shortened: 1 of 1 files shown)


jlost:
If this is going to be fully correct for Webpack 2, "loaders" should also be changed to "rules"...

RyanCavanaugh:

PR #555, opened 2017-04-18
Another way to explain the pros and cons of shadowing

vitoc:
I think it's better to first explain how TypeScript's handling of shadowing helps prevent bugs before going on to explain how shadowing can introduce certain bugs in the event of accidental shadowing.

Also includes a camelization fix.


DanielRosenwasser:
Hey, I'm sorry I missed this! Could you update to the current branch and also keep each sentence on its own line? It makes reviewing easier.

RyanCavanaugh:

PR #556, opened 2017-04-19
Clarifies Mapped Types and additional members

dallonf:
Resolves https://github.com/Microsoft/TypeScript-Handbook/issues/540.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -796,22 +796,20 @@ type PersonPartial = Partial<Person>;
 type ReadonlyPerson = Readonly<Person>;
 ```
 
-Note that this syntax describes a type rather than a member. You **cannot** add additional members directly to a mapped typed:
+Note that this syntax describes a type rather than a member. If you want to add additional members, use an intersection type:
 
 ```ts
+// Use this
+type PartialWithNewMember<T> = {
+  [P in keyof T]?: T[P];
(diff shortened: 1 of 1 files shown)


pages/Advanced Types.md
```diff
@@ -796,6 +796,24 @@ type PersonPartial = Partial<Person>;
...
+
+```ts
+type PartialWithNewMember<T> = {
+  [P in keyof T]?: T[P];
+  // This is an error
+  newMember: boolean;
+}
+```
+
+Although you can use union types for a similar effect:
```
mhegazy: intersection and not union

pages/Advanced Types.md
```diff
@@ -796,6 +796,24 @@ type PersonPartial = Partial<Person>;
 type ReadonlyPerson = Readonly<Person>;
 ```
 
+Note that this syntax describes a type rather than a member. You **cannot** add additional members directly to a mapped typed:
```
mhegazy: Instead of saying this is an error, I would say if you want to add additional members use intersection types, and here is an example.

dallonf:
Thanks for the feedback. I've made some changes, does that read better?

RyanCavanaugh:
@dallonf need CLA signing before I can merge this

RyanCavanaugh:
Better late than never

PR #560, opened 2017-04-20
Suggestion to change 'type' to 'destructuring' to describe function declaration with destructuring and default values

### LINKED ISSUE #558: The term 'type' used in "Function declarations" in Variable Declaration is confusing
In Variables Declaration, there's the following:

"
But specifying defaults is more common for parameters, and getting defaults right with destructuring can be tricky. First of all, you need to remember to put the _type_ before the default value.

```
function f({ a, b } = { a: "", b: 0 }): void {
    // ...
}
f(); // ok, default to { a: "", b: 0 }

```
"
I feel like the term _type_ is confusing. Clearly there's no type (as I understood it) here before the default value. The type is inferred from the assignment? What's before the default values are the variables assigned to in the deconstruction? 

It may also be good to have a short note on type inference.

vitoc:
Fixes #558

### DIFF
--- a/pages/Variable Declarations.md
+++ b/pages/Variable Declarations.md
@@ -578,7 +578,7 @@ function f({ a, b }: C): void {
 ```
 
 But specifying defaults is more common for parameters, and getting defaults right with destructuring can be tricky.
-First of all, you need to remember to put the destructuring before the default value.
+First of all, you need to remember to put the pattern before the default value.
 
 ```ts
 function f({ a, b } = { a: "", b: 0 }): void {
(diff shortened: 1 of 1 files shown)


pages/Variable Declarations.md
```diff
@@ -578,7 +578,7 @@ function f({ a, b }: C): void {
 ```
 
 But specifying defaults is more common for parameters, and getting defaults right with destructuring can be tricky.
-First of all, you need to remember to put the type before the default value.
+First of all, you need to remember to put the destructuring before the default value.
```
mhegazy: how about "pattern" instead?

vitoc: Better! Thanks and updated!

PR #563, opened 2017-04-24
.js comment in Declaration Merging should be .ts (Issue #562)

### LINKED ISSUE #562: .js comment in Declaration Merging should be .ts
The following snippet in "Module Augmentation" section:

```
// observable.js
export class Observable<T> {
    // ... implementation left as an exercise for the reader ...
}

// map.js
import { Observable } from "./observable";
Observable.prototype.map = function (f) {
    // ... another exercise for the reader
}
```

.js should be .ts

vitoc:
Fixes #562

### DIFF
--- a/pages/Declaration Merging.md
+++ b/pages/Declaration Merging.md
@@ -248,12 +248,12 @@ Although JavaScript modules do not support merging, you can patch existing objec
 Let's look at a toy Observable example:
 
 ```js
-// observable.js
+// observable.ts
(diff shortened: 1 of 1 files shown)


mhegazy:
The use of `.js` here is intentional. it describes a code pattern in JS and then how it can be modeled in a TS.

vitoc:
Ah, thanks. That clarifies.

PR #597, opened 2017-05-21
Update Integrating with Build Tools.md to use latest WebPack configuration

thomasclaudiushuber:
Update webpack.config.js code, as element module{ loaders:[...]} does not exist anymore in the latest webpack version. It's now module{rules[...]}. See also basic webpack.config.js on https://webpack.js.org/guides/webpack-and-typescript/.
Also added a link to that webpack.js.org-pack that shows the typescript integration

Fixes #

### DIFF
--- a/pages/Integrating with Build Tools.md
+++ b/pages/Integrating with Build Tools.md
@@ -156,20 +156,26 @@ module.exports = {
     },
     resolve: {
         // Add '.ts' and '.tsx' as a resolvable extension.
-        extensions: ["", ".webpack.js", ".web.js", ".ts", ".tsx", ".js"]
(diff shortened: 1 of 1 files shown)


thomasclaudiushuber:
Sorry, I've seen there's a similar, not yet merged PR here: https://github.com/Microsoft/TypeScript-Handbook/pull/514

thomasclaudiushuber:
There seems to be a PR for this topic, which is already out of date https://github.com/Microsoft/TypeScript-Handbook/pull/510

mhegazy:
@DanielRosenwasser mind taking a look.

RyanCavanaugh:

PR #604, opened 2017-06-13
Try to clarify what an "own, enumerable property" is

brettcannon:
If you're not familiar with JavaScript, what an "own, enumerable property" is not obvious. Specifying that this has to do with the object being spread clarifies (somewhat) what is being discussed.


mhegazy:
not sure this rely changes anything. i would recommend adding a link to [MDN's article about ownership and enumrability in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties) instead.

brettcannon:
Added the link as requested.

PR #623, opened 2017-07-11
Link to external article explaining this is broken

schatekar:
The link to Yahuda Katz's article explaining this seems to broken. I am not sure if this is temporary but I have managed to find a Google's cached copy of the article.

Fixes #


mhegazy:
the link seems to be working fine for me. keeping it unless other users are running into the same issue.

PR #625, opened 2017-07-19
add export for module augmentation

macjohnny:
in order to prevent the module augmentation being discarded by e.g. tree shaking

Fixes #

### DIFF
--- a/pages/Declaration Merging.md
+++ b/pages/Declaration Merging.md
@@ -268,7 +268,7 @@ You can use module augmentation to tell the compiler about it:
 // map.ts
 import { Observable } from "./observable";
 declare module "./observable" {
-    interface Observable<T> {
+    export interface Observable<T> {
         map<U>(f: (x: T) => U): Observable<U>;
     }
 }


mhegazy:
> in order to prevent the module augmentation being discarded by e.g. tree shaking

not sure i understand what this means. the `export` is optional here, since it is inside an ambient declaration, and all ambient declarations are assumed to be exported. 
This is an interface, and should not have any tree shaking implications.

macjohnny:
@mhegazy we faced a problem when using module augmentation without redeclaring the `export`, where the compiler did not recognized the merged declaration. We were able to solve it with the `export` keyword.
Moreover, for unit tests with karma, in the file augmenting a module, we needed to explicitly re-export the interface and import it in a test in order for the compiler to know about the augmented interface and prevent tree shaking. 
However, if you think this is not directly related to TypeScript, but to the deployment software (angular CLI, karma, etc.), you can close the pull request.

mhegazy:
This does not sound like a documentation issue but more of a bug. feel free to file a new issue with more details on https://github.com/Microsoft/TypeScript

PR #632, opened 2017-08-11
Document `--preserveSymlinks` compiler option

andy-ms:
Added this in Microsoft/TypeScript#16661.
Do we have a separate way of adding this to the JSON schema for `tsconfig.json`? @DanielRosenwasser


DanielRosenwasser:
@paulvanbrenk usually updates the JSON schema in [schemastore](http://schemastore.org/json/).

paulvanbrenk:
Do we have other compiler options which are not part of the schema?

@andy-ms I can show you how to update the schema in the [Schemastore repo](https://github.com/schemastore/schemastore)

andy-ms:
Also: Microsoft/TypeScript#17099

PR #651, opened 2017-09-27
React & React-Dom file location changed

arthurchipdean:
Update the include path in the script tags to include the right library location for react & react-dom.

Fixes #


DanielRosenwasser:
Thanks! Should be updated relatively soon.

PR #654, opened 2017-09-29
Make a piece of the example more clear

Moonshile:
Update the description of type `void` and add another example.

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -184,10 +184,11 @@ function warnUser(): void {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pages/Basic Types.md
```diff
@@ -184,10 +184,11 @@ function warnUser(): void {
 }
 ```
 
-Declaring variables of type `void` is not useful because you can only assign `undefined` or `null` to them:
+Declaring variables of type `void` is not useful because you can only assign `undefined`, `null` or a void function call to them:
```
mhegazy: a `void` function returns `undefined` at runtime. not sure this addition adds anything to the meaning.

PR #662, opened 2017-10-11
Remove "algebraic data types"

bsima:
This is incorrect. A tagged union is one way to implement sum types, which is a kind of algebraic data type. However, product types (another common ADT) is not possible in TypeScript.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -543,7 +543,7 @@ Much of the time when we talk about "singleton types", we're referring to both e
 
 # Discriminated Unions
 
-You can combine singleton types, union types, type guards, and type aliases to build an advanced pattern called *discriminated unions*, also known as *tagged unions* or *algebraic data types*.
+You can combine singleton types, union types, type guards, and type aliases to build an advanced pattern called *discriminated unions*, also known as *tagged unions*.
 Discriminated unions are useful in functional programming.
 Some languages automatically discriminate unions for you; TypeScript instead builds on JavaScript patterns as they exist today.
 There are three ingredients:


heejongahn:
+1 for more correct documentation. However, TypeScript supports a tuple type. Isn't that a product type?

DanielRosenwasser:
Tuple types and record types (i.e. objects) are both product types, so I'm not even sure if the pedantry here is correct. ;)

bsima:
The pedantry is correct because a tagged union is an ADT but not all ADTs are tagged unions. The documentation will be more understandable with concrete descriptions and explanations, not with grand claims about general support. Let me explain with an example.

In Typescript this is a union type:

```typescript
type Thing =
  | Widget
  | Gadget;
```

whereas in other languages such as Elm a similar syntax is a sum type. In order to get the effect of sum types with tagged unions, you have to manually specify the tags; in other ML-derived languages the compiler does this for you. (Note that tagged unions and sum types are not really the same thing. https://github.com/Microsoft/TypeScript/issues/186 was opened asking for sum types and closed with the addition of tagged unions, to some users' dismay.)

This is a source of confusion for people coming from other ML-derived languages. I'm trying to make this clearer for people coming from these languages. Saying that TS supports ADT's implies that the compiler handles this for you, but in reality TS has as much support for ADT's as Lisp (yeah you can implement these types with raw lists and macros, but you need to put in a lot more work than you do in Haskell/Elm). 

Maybe it would make more sense to describe TS types in the context of ML, and give common alternatives and patterns, rather than just a generalized "TS supports ADTs" claim.

PR #666, opened 2017-10-20
--lib defaults are a list of libraries

mikemaccana:
--lib defaults are a list of libraries, not a single library

<!--


mikemaccana:
Mua PR 666 🤘😈

mikemaccana:
Thanks! uoy sevol natas.

PR #667, opened 2017-10-22
add description of declaration-including package to Consumption.md

noymer:
in Declaration Files `Publishing` page, 
it is written that you can include declaration files in npm package of your own.
So in consumption phase, I **assume** if the npm package has already include declaration files 

For example `reactjs/redux` has `index.d.ts`, so this command will not be needed.

```
npm install --save @types/redux 
```

https://github.com/reactjs/redux/blob/master/index.d.ts

In case my assumption is true,
It is better to add the description.


noymer:
Thank you!

PR #672, opened 2017-10-25
Webpack updates

jamierytlewski:
Fixes #


RyanCavanaugh:

PR #676, opened 2017-10-27
Add --strict compiler option in msbuild

styfle:
The `--strict` flag was added in https://github.com/Microsoft/TypeScript/pull/14486


styfle:
This flag was recommend by @mhegazy

mhegazy:
thanks

PR #678, opened 2017-10-31
Fix syntax error: use `function` instead of `const` in declaration

joanrieu:
This PR fixes a small syntax error / typo where a function was declared with the wrong keyword.

Fixes Microsoft/TypeScript#14047


mhegazy:
thanks!

Issue #670, opened 2017-10-23
Small example with inheritance


kodenix:
Hello, i learning the typescript tutorial and the example with inheritance is confusing but not exist a example using the method of mother class directily without defined other method with the same name. This can cause confusing a otros, i suggest show a example simple o f inherit if is possible.

Example:

```
class Mother {
protected talk(){
console.log('Hello');
}
}

class Child extends Mother {

}

let child=new Child();
child.talk();
```
Best reggards

DanielRosenwasser:
If I understand the request correctly, I think you're right. The handbook seems to assume people are already familiar with inheritance and overriding.

PR #679, opened 2017-10-31
Add simpler example of inheritance

joanrieu:
This PR adds a small and simple example of inheritance to introduce the concept before showing advanced use cases.

Solves issue #670.


DanielRosenwasser:
Thanks for the pull request @joanrieu!

PR #681, opened 2017-11-06
clarification for triple-slash-reference-types

dennispg:
Currently the documentation states that triple slash reference types will only resolve with `@types` packages, but this is not correct.


DanielRosenwasser:
Good catch, and thank you for the contribution!

PR #682, opened 2017-11-07
Remove inaccurate section about type alias names

jayphelps:
Likely true at one point, but right now aliases actually do create a name that is used for Intellisense and errors. However, if you make aliases of another alias it just uses the first alias name; this has been marked as a bug, but it's possibly intended behavior: https://github.com/Microsoft/TypeScript/issues/19198

[Playground Demo](http://www.typescriptlang.org/play/#src=type%20Alias%20%3D%20%7B%20num%3A%20number%20%7D%0D%0Ainterface%20Interface%20%7B%0D%0A%20%20%20%20num%3A%20number%3B%0D%0A%7D%0D%0Adeclare%20function%20aliased(arg%3A%20Alias)%3A%20Alias%3B%0D%0Adeclare%20function%20interfaced(arg%3A%20Interface)%3A%20Interface%3B%0D%0A%0D%0Aconst%20alias%3A%20Alias%20%3D%20%7B%20num%3A%201%20%7D%3B%20%0D%0Aconst%20foo%3A%20string%20%3D%20aliased(alias)%3B)

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -447,22 +447,7 @@ type Yikes = Array<Yikes>; // error
 
 ## Interfaces vs. Type Aliases
 
-As we mentioned, type aliases can act sort of like interfaces; however, there are some subtle differences.
-
-One difference is that interfaces create a new name that is used everywhere.
-Type aliases don't create a new name &mdash; for instance, error messages won't use the alias name.
-In the code below, hovering over `interfaced` in an editor will show that it returns an `Interface`, but will show that `aliased` returns object literal type.
-
-```ts
-type Alias = { num: number }
-interface Interface {
-    num: number;
-}
-declare function aliased(arg: Alias): Alias;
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Well, there's no guarantee you'll always get the alias name. It's a best-effort thing. @mhegazy  can fill in more here.

Probably the real inaccuracy here is that type aliases *can* be extended if they refer to object types or intersections of object types, but they can't be reopened.

mhegazy:
type aliases are not types, but rather aliases. in other words, the compiler do not see them as entities per se, but only sees their expanded form. that is why a type alias can not reference itself without an indirection for example. 
for error reporting purposes, the compiler tries to associate the name of the alias to the expanded type it aliases. this works if the type was not referenced before, but there are no guarantees that this would work for the general case..

jayphelps:
Okiedoke. Sounds like the docs are as intended. Thanks!

PR #683, opened 2017-11-10
Typo fix

Hannes-Magnusson-CK:
We are talking about `const` here, not `var`.
We've already spent much time talking about why `var` isn't a good idea


mhegazy:
The use of `var` here is intentional.. `let` and `var` is shorter than `const`, and that is what this statement is saying.

Hannes-Magnusson-CK:
The statement (as-is) is saying that many people prefer writing `var` because its shorter to write the `const` which was suggested by the previous paragraph.
Its pretty ambiguous sentence though, and could just be removed as it doesn't really add any value

mhegazy:
I am fine removing the whole sentence.

PR #684, opened 2017-11-12
Method decorator factory is missing return type

robsonrosa:
The example supplied in Method Decorator is not working in TS `2.6` because a return type is expected in the method decorator factory and it must be `any` or `void`.

Fixes #

### DIFF
--- a/pages/Decorators.md
+++ b/pages/Decorators.md
@@ -48,7 +48,7 @@ A *Decorator Factory* is simply a function that returns the expression that will
 We can write a decorator factory in the following fashion:
 
 ```ts
-function color(value: string) { // this is the decorator factory
+function color(value: string): any { // this is the decorator factory
     return function (target) { // this is the decorator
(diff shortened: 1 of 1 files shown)


pages/Decorators.md
```diff
@@ -203,6 +203,10 @@ If the method decorator returns a value, it will be used as the *Property Descri
 
 > NOTE&emsp; The return value is ignored if your script target is less than `ES5`.
 
+A return type is expected at the method decorator factory and it must be `any` or `void`.
```
mhegazy: this is not correct.. a return type is inferred, and here it is a function. I think you are running into other issues, adding  `any` just shuts off checking.

mhegazy:
this is a valid decorator:

```ts
function color(value: string) { // this is the decorator factory
    return function (target) { // this is the decorator
        // do something with 'target' and 'value'...
    }
}

@color("red")
class C { 
    p: number;
}
```

PR #685, opened 2017-11-14
Remove ambiguous sentence about let/var/const brevity

Hannes-Magnusson-CK:
Remove ambiguous sentence, and push the handbook usage to the end for improved flow.

See also https://github.com/Microsoft/TypeScript-Handbook/pull/683#issuecomment-344011080


mhegazy:
thanks!

PR #691, opened 2017-12-06
add missing dep

noscripter:
Fixes # add missing dependency 'babel-core'


DanielRosenwasser:
We'll eventually need to update these for the scoped package names, but this looks good. Thanks!

PR #693, opened 2017-12-13
Underscore for backing field of get/set accessor

devuxer:
Properly document why underscore is used for backing field when get/set accessor is used.

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -344,6 +344,7 @@ First, accessors require you to set the compiler to output ECMAScript 5 or highe
 Downlevelling to ECMAScript 3 is not supported.
 Second, accessors with a `get` and no `set` are automatically inferred to be `readonly`.
 This is helpful when generating a `.d.ts` file from your code, because users of your property can see that they can't change it.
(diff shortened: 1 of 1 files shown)


mhegazy:
not sure i see much value in the additional sentence.

devuxer:
I had to do some detective work, reading through GitHub issues, to figure out why underscores are generally not recommended except in this one case. I was hoping to save others the same hassle. Coming from C#, where there's a history of using underscore for all private fields (then a more recent change in guidelines suggesting underscores not be used), I was definitely confused when to use underscores and when not to and why.

mhegazy:
In general this is not a pattern we recommend, having an accessor with a packing private field. we recommend instead using methods. 
as for `_` we recommend not using them altogether.  it might be better to change the example to not use `_`.

devuxer:
If it's the case that you don't recommend `_` in any situation, then I entirely agree that the example should be changed (and my pull request should not be merged). (I'm not quite I sure understand what you mean by "methods", though--aren't accessors methods?)

RyanCavanaugh:

PR #698, opened 2017-12-18
Add select method to TextBox class in interfaces example

bward:
Fixes [#20701](https://github.com/Microsoft/TypeScript/issues/20701) on the main Typescript repository


mhegazy:
thanks!

PR #704, opened 2017-12-29
Fix broken relative links

mike-north:
Github is unhappy with spaces in markdown link URLs. They must be manually encoded


DanielRosenwasser:
*sigh* this will actually break [this thing](https://github.com/DanielRosenwasser/JekyllMarkdownLinkConverter) so we should probably just modify the filenames to use dashes and call it a day.

mike-north:
@DanielRosenwasser Works for me. I used this regex to avoid missing anything
```
[\s]+[A-Za-z]+.md\)
```

RyanCavanaugh:

PR #705, opened 2018-01-01
An import path cannot end with a '.tsx' extension

ethanx94:
This will fix an error. I researched other solutions that might make modifications to the tsconfig.json but this seemed to be the most straightforward way.

Fixes #


mhegazy:
thanks!

PR #709, opened 2018-01-03
Also list --noErrorTruncation as a supported flag

Vinnl:
To find out how to prevent my error messages from getting truncated, I had to dig through source code that happened to be the first hit after a search. There might be a reason this option is not listed in the documentation, but if the reason was that it was simply forgotten, this PR fixes that :)


mhegazy:
thanks!

PR #710, opened 2018-01-03
Fix order of sections

valera-rozuvan:
In the left panel (table of contents menu), the section "Do”s and “Don’t”s" comes after the section "By Example".

See https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html to see what I mean, and take a note of the left panel.


mhegazy:
thanks!

PR #716, opened 2018-01-24
Fix typo in optional interface property example

Cludch:
As I was reading the example code, I stumpled upon this one line, which does not seem logical to me.
I believe it is supposed to be the way I changed it.


mhegazy:
thanks!

Issue #366, opened 2016-08-13
Suggestion for update examples for Interface page


yuit:
_This is an [issue ](https://github.com/Microsoft/TypeScript/issues/10301)moved from TypeScript Repo_

Handbook link: https://www.typescriptlang.org/docs/handbook/interfaces.html

In the last topic, Interfaces Extending Classes, the extends in the example code differ from what the explanation below tells.

Here is my suggestion for a new code example:

``` ts
class Control {
    private state: any;
}

interface SelectableControl extends Control {
    select(): void;
}

class Button extends Control implements SelectableControl {
    select() { }
}

class TextBox extends Control implements SelectableControl {
    select() { }
}

class Image extends Control {
}

class Location extends Control {
    select() { }
}
```

vovanmozg:
I also think that an example in the last topic looks incomplete.

I suggest my version of code https://github.com/Microsoft/TypeScript-Handbook/pull/637

```ts
class Control {
    private state: any;
}

interface SelectableControl extends Control {
    select(): void;
}

class Button extends Control implements SelectableControl {
    select() { }
}

class TextBox extends Control {
    
}

// Error: Property 'state' is missing in type 'Image'.
class Image implements SelectableControl {
    select() { }
}

class Location {
    
}
```

vovanmozg:
@roblperry thank you for resuming the discussion. My suggestion was wrong. And the documentation example was wrong too.

I looked the changes history of interfaces.html. It seems that the original version was correct, but as a result of several edits, it began to be misleading (see https://github.com/Microsoft/TypeScript-Handbook/commit/25ae2209d442d095f7a2c51e8a54948b2ea26612#diff-0d122fa6d7fac71313bb600040376e83 by @sandersn ).

Suggestion:
1) to restore the code version of the creator

```ts
class Control {
    private state: any;
}

interface SelectableControl extends Control {
    select(): void;
}

class Button extends Control {
    select() { }
}

class TextBox extends Control {
    select() { }
}

class Image extends Control {
}

class Location {
    select() { }
}
```

2) add some text to the last paragraph:
`Image` has not `select()` method. `Location` has not private property `state`;

3) add additional example
```ts
let someVar: SelectableControl;

// Acceptable, because The TextBox is a descendant of the Control class
someVar = new TextBox();

// Type 'Image' is not assignable to type 'SelectableControl'.  Property 'select' is missing in type 'Image'.
someVar = new Image();
 
[...]

PR #717, opened 2018-01-27
Addressed issue #366 by making TextBox implement select(): void via e…

roblperry:
Addressed issue #366 by making TextBox implement select(): void via extending a class the implements it, as opposed to implementing it direction as Button does.


mhegazy:
The issue seems to be have been fixed by https://github.com/Microsoft/TypeScript-Handbook/pull/698. I do not think this change is required.

PR #720, opened 2018-02-09
Improved Mixins Helper Function to support getters & setters

MRamos62786:
I was having trouble following along with the mixins help function because it was actually calling a getter when it's suppose to only be copying other properties from one prototype to another. I've update the sample code and have confirmed it compiles in the latest version of typescript.


MRamos62786:
I've been thinking about this some more, and it looks like this could be improved further. It looks like this helper function will just overwrite any overrides a class that is implementing the mixin has. I'm thinking is should use Object.getOwnPropertyDescriptor() on the derived class to see whether or not the property exists prior to overwriting it. Thoughts?

PR #735, opened 2018-03-19
Add a step to avoid a compiler error in VSCode

epicbugs:
Add a step to "TypeScript in 5 minutes" for avoiding a cryptic compiler/intellisense error in VSCode (see https://stackoverflow.com/q/44192541/426379).

### DIFF
--- a/pages/tutorials/TypeScript in 5 minutes.md
+++ b/pages/tutorials/TypeScript in 5 minutes.md
@@ -35,7 +35,13 @@ document.body.innerHTML = greeter(user);
 We used a `.ts` extension, but this code is just JavaScript.
 You could have copy/pasted this straight out of an existing JavaScript app.
 
-At the command line, run the TypeScript compiler:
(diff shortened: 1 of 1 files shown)


mhegazy:
do not see why this is required. opening that file alone in VSCode should just work with no errors. please share the full context for us to reproduce the issue.

epicbugs:
@mhegazy Thanks for looking at this. Here's the context:

1. Create folder "test"
2. Launch VS Code and open folder "test" using "File -> Open Folder..."
3. Copy-paste second code example from tutorial and save it as "greeter.ts"
4. Open VS Code terminal and execute "tsc greeter.ts"
5. Open the compilation result "greeter.js" in another VS Code tab
6. Go back to "greeter.ts", add a newline (or whatever change that keeps identifier names the same)
7. Observe VS Code reporting problems
8. Open VS Code terminal and execute "tsc --init"
9. Observe the reported problems disappear

RyanCavanaugh:
This change makes the rest of the example kind of nonsensical, and we generally don't want to have workarounds for other tools confusing the Handbook content.

PR #738, opened 2018-03-24
Correct a MSBuild configuration comment to have a consistent instruction

Bartizan:
Corrected a MSBuild configuration comment to have a consistent instruction.


mhegazy:
thanks!

PR #746, opened 2018-04-06
Fixed typo

aladdine:

### DIFF
--- a/pages/Interfaces.md
+++ b/pages/Interfaces.md
@@ -223,7 +223,7 @@ interface SearchFunc {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pages/Interfaces.md
```diff
@@ -223,7 +223,7 @@ interface SearchFunc {
 }
 ```
 
-Once defined, we can use this function type interface like we would other interfaces.
+Once defined, we can use this function type interface like we would in other interfaces.
```
mhegazy: do not think this a typo. it just says that this function type is just like any other interface..

PR #748, opened 2018-04-09
Improved code in 'Modules'

the1812:
Use `ProgrammerCalculator.digits.length` in place of `16` (in class `ProgrammerCalculator`).


mhegazy:
thanks!

PR #753, opened 2018-04-24
Correct syntax typo for import statement

tschaub:
This corrects the syntax for the example import statement.


RyanCavanaugh:
The suggested change isn't correct for the context it's written in.

PR #764, opened 2018-05-24
[Proposition] replace pluck by props in Advanced types > Index type

ArnaudBertrand:
Change #

The guilty:

```
function pluck(o, names) {
     return names.map(n => o[n]);
}
```

The name pluck can be confusing.
In ramda or underscore pluck is passing a property name and map each object of the array to that property name.
In lodash it's just a map.
In RxJS you map to emitted value to a property (each argument being a new depth).
etc....
Here it's again different (the array contains the props of the object you want)

If we can just name this "props" or what suits you it would be less confusing and we could read the document without needing to reading the implementation of the function.

What do you think?


RyanCavanaugh:

PR #767, opened 2018-05-29
The target of TypeScript is es6

raoenhui:
The target of TypeScript is es6. If target is es5,have error message "ERROR in [at-loader] ./node_modules/@types/node/index.d.ts"
image


mhegazy:
@DanielRosenwasser please review this change

PR #770, opened 2018-05-31
Replace alert function with console.log where applicable

oboukli:
Fixes:

- Replace alert function with console.log where applicable
- Fix minor style issues


mhegazy:
thanks!

PR #773, opened 2018-06-05
Fixes small grammatical error

iAmWillShepherd:
I did not create an issue for this and I'm pretty sure one doesn't exist. I was reading the docs and noticed a small error, so I decided to do the needful and add **the** missing article 😏


mhegazy:
thanks!

PR #777, opened 2018-06-07
Fixed destructing function declaration example

### LINKED ISSUE #729: Destructing section is not pragmatic
From @rauschma:

Destructuring in function declarations seems wrong: https://t.co/Qg7HR4NQgx

You’d never do:
f({ a, b } = { a: "", b: 0 })

Instead:
f({ a="", b=0 } = {})
You can even omit the ={} at the end in many cases.

It’s a shame that the manual warns about this pattern which is really useful.

Vimal-Raghubir:
Fixes #729. 

The change involves removing the old example which was deemed wrong by @DanielRosenwasser in the issue above, and replaced it with the same user's suggested fix.

### DIFF
--- a/pages/Variable Declarations.md
+++ b/pages/Variable Declarations.md
@@ -583,7 +583,7 @@ First of all, you need to remember to put the pattern before the default value.
 function f({ a="", b=0 }): void {
     // ...
 }
-f(); // ok, default to { a: "", b: 0 }
+f();
 ```
 
 > The snippet above is an example of type inference, explained later in the handbook.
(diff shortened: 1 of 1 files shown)


pages/Variable Declarations.md
```diff
@@ -580,10 +580,10 @@ But specifying defaults is more common for parameters, and getting defaults righ
 First of all, you need to remember to put the pattern before the default value.
 
 ```ts
-function f({ a, b } = { a: "", b: 0 }): void {
+function f({ a="", b=0 }): void {
```
mhegazy: you also need `= {}` for it to be callable with no arguments

mhegazy:
thanks!

PR #778, opened 2018-06-07
Updated React and Webpack.md to fix the error thrown by tslint

yvrkarthik:
Hi there,

I have added the typedef to the const in the stateless functional component code on line 125.  tslint was complaining about the expected typedef for the const variable .

image

Thanks
Rajiv


mhegazy:
I do not think every stateless component needs to have a type declaration. this seems excessive. I would say either disable the lint rule, or file a bug on the tslint side.

yvrkarthik:
@mhegazy sounds good 👍 .

PR #787, opened 2018-07-12
Add new section "Conditional Types". Fixed #747

### LINKED ISSUE #747: Documentation for Conditional Types?
It doesn't seem that Conditional Types are described in the [TypeScript Handbook](https://www.typescriptlang.org/docs/index.html). I would expect to find them under [Advanced Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html) (or on their own page) - no other page seems to fit - but a quick search on that page for "conditional", "extends", "?", and similar keywords yields nothing relevant.

It seems that they are currently only described in the [TypeScript 2.8 Release Notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html), which is kind of a bummer if you haven't memorized the exact version that added the feature :)

Fortunately, the Release Notes are pretty descriptive, so they could probably be copied from there into Advanced Types with minimal edits. I could take a stab at doing that in a PR - just wanted to check first that there's no deeper reason why they haven't been documented or if it actually is somewhere else that I didn't think to look.

zhongsp:
Add new section "Conditional Types". Fixed #747

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -938,14 +938,19 @@ T extends U ? X : Y
 The type above means when `T` is assignable to `U` the type is `X`, otherwise the type is `Y`.
 
 A conditional type `T extends U ? X : Y` is either *resolved* to `X` or `Y`, or *deferred* because the condition depends on one or more type variables.
-Whether to resolve or defer is determined as follows:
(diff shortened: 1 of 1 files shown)


mhegazy:
thanks!

@DanielRosenwasser any reservations before we merge?

DanielRosenwasser:
I made some changes, but thanks for starting this off!

PR #788, opened 2018-07-12
Project References

RyanCavanaugh:
Adds documentation for project references and `tsc -b`

### DIFF
--- a/pages/Project References.md
+++ b/pages/Project References.md
@@ -1,0 +1,203 @@
+# Introduction
+
+Project references, introduced in TypeScript 3.0, allow you to structure your TypeScript programs into smaller pieces. By doing this, you can greatly improve build times, enforce logical separation between components, and organize your code in new and better ways.
+
+A new mode for for the command-line compiler, `tsc --build`, works with project references to orchestrate builds when working with multi-project solutions. You can also use existing build tools like `gulp` or `webpack` with project references.
+
+## Project References
+
+We refer to each `tsconfig.json` file as a "project". A project reference simply says "This project *depends on* this other project".
+
+You specify these references in a new top-level property in `tsconfig.json`, `references`. It's an array of objects that specifies projects to reference:
+```json5
+{
+    "compilerOptions": {
+        // The usual
+    },
+    "references": [
+        { "path": "../src" }
+    ]
+}
+```
+
+The `path` property of each reference can point to a directory containing a `tsconfig.json` file, or to the config file itself (which may have any name). This is the same convention used for running `tsc -p` / `tsc --project`.
+
+When you reference a project, new things happen:
+ * Importing modules from a referenced project will instead load its *output* declaration file (`.d.ts`)
+ * If the referenced project produces an `outFile`, the output file `.d.ts` file's declarations will be visible in this project
+ * Build mode (see later on this page) will automatically build referenced projects if needed
+
+By separating into multiple projects, you can greatly improve the speed of typechecking and compiling, reduce memory usage when using an editor, and improve enforcement of the logical groupings of your program.
+
+## An Example Project
+
+Let's look at a fairly normal program and see how project references can help us better organize it.
+Imagine you have a project with two modules, `converter` and `units`, and a corresponding test file for each:
+```
+/src/converter.ts
+/src/units.ts
+/test/converter-tests.ts
+/test/units-tests.ts
+/tsconfig.json
+```
+
+The test files import the implementation files and do some testing:
+```
+// converter-tests.ts
+import * as converter from "../converter";
+
+assert.areEqual(converter.celsiusToFahrenheit(0), 32);
+```
+
+Previously, this structure was rather awkward to work with if you used a single tsconfig file:
+ * It was possible for the implementation files to import the test files
+ * It wasn't possible to build `test` and `src` at the same time without having `src` appear in the output folder name, which you probably don't want
+ * Changing just the *internals* in the implementation files required *typechecking* the tests again, even though this wouldn't ever cause new errors
+ * Changing just the tests required typechecking the implementation again, even if nothing changed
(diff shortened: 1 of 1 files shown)


pages/Project References.md
```diff
@@ -0,0 +1,203 @@
...
+
+The test files import the implementation files and do some testing:
+```
+// converter-tests.ts
+import * as converter from "../converter";
+
+assert.areEqual(converter.celsiusToFahrenheit(0), 32);
+```
+
+Previously, this structure was rather awkward to work with if you used a single tsconfig file:
```
mhegazy: since this is a handbook page, and not a release blog, i would not use "previously". i would just say "Building as a single project would"

pages/Project References.md
```diff
@@ -0,0 +1,203 @@
...
+
+One of the main reasons to use project references is to improve build speed. It's important to understand how this happens and be aware of the implications.
+
+When loading types from a referenced project, types always come from the generated `.d.ts` file instead of an implementation (`.ts`) file. This is much faster because all type inference has already taken place, there's no need to check for errors in statements or expressions, and the amount of memory and disk I/O needed is also lower. However, this means that the referenced project does need to be built before the referencing project can understand its own types. If you open a multi-project solution in an editor, you may see a large number of errors caused by this until the referenced projects get built. We're working on a behind-the-scenes .d.ts generation process that should be able to mitigate this, but for now we recommend informing developers that they should build after cloning.
+
+Because `.d.ts` files themselves don't *generate* new `.d.ts` files, you'll want to write any hand-authored types in a `.ts` file instead of a `.d.ts` file, even if you don't intend to include any executable statements in the file.
```
mhegazy: i do not think this is what we want ppl to be doing.. they just need to make the .d.ts files in a shared location. and include them in both projects.

pages/Project References.md
```diff
@@ -0,0 +1,203 @@
...
+## `composite`
+
+Referenced projects must have the new `composite` setting enabled.
+This setting is needed to ensure TypeScript can quickly determine where to find the outputs of the referenced project.
+Enabling the `composite` flag changes a few things:
+ * The `rootDir` setting, if not explicitly set, defaults to the directory containing the `tsconfig` file.
+ * All implementation files must be matched by an `include` pattern or listed in the `files` array. If this constraint is violated, `tsc` will inform you which files weren't specified.
+ * `declaration` defaults to `true`
+
+These changes won't affect many projects in a substantial way. We recommend turning on `composite` in most projects, even if they're not using project references yet.
```
mhegazy: should we add it to `tsc --init`?

pages/Project References.md
```diff
@@ -0,0 +1,203 @@
...
+
+# More Options for Project References
+
+## `declarationMaps`
```
mhegazy: i think we should make this on by default if `composite` is set. this way ppl do not have to worry about it.

pages/Project References.md
```diff
@@ -0,0 +1,203 @@
...
+B     C
+ ^   ^
+  \ /
+   D
+```
+It's important in this situation to not prepend at each reference, because you'll end up with two copies of `A` in the output of `D` - this can lead to unexpected results.
+
+# Build Mode for TypeScript
+
+A long-awaited feature is smart incremental builds for TypeScript projects. In 3.0 you can use the `--build` flag with `tsc`. This is effectively a new entry point for `tsc` that behaves more like a build orchestrator than a simple compiler. `--build` changes the meaning of other flags, so it must be the first argument to `tsc` if it's being used.
```
mhegazy: do not think you need the `long-awaited` part, it will not make much sense when someone looks at the handbook a year from now.

pages/Project References.md
```diff
@@ -0,0 +1,203 @@
...
+There are also some flags specific to `tsc -b`:
+ * `--verbose`: Prints out verbose logging to explain what's going on (may be combined with any other flag)
+ * `--dry`: Shows what would be done but doesn't actually build anything
+ * `--clean`: Deletes the outputs of the specified projects (may be combined with `--dry`)
+ * `--force`: Act as if all projects are out of date
+ * `--watch`: Watch mode (may not be combined with any flag except `--verbose`)
+
+## Caveats
+
+Normally, `tsc` will produce outputs (`.js` and `.d.ts`) in the presence of syntax or type errors, unless `noEmitOnError` is on. Doing this in an incremental build system would be very bad - if one of your out-of-date dependencies had a new error, you'd only see it *once* because a subsequent build would skip building the now up-to-date project. For this reason, `tsc -b` effectively acts as if `noEmitOnError` is enabled for all all projects.
```
mhegazy: i would just say if one project has any errors, the compiler will not attempt building referencing projects. (people do not need to know what the `noEmitOnError` behavior is to understand that part). you can always have a reference to it at the end.

pages/Project References.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Structuring for outFiles
+
+Layout for compilations using `outFile` is more flexible because relative paths don't matter as much. One thing to keep in mind is that you'll generally want to not use `prepend` until the "last" project - this will improve build times and reduce the amount of I/O needed in any given build. The TypeScript repo itself is a good reference here - we have some "library" projects and some "endpoint" projects; "endpoint" projects are kept as small as possible and pull in only the libraries they need.
+
+## Structuring for monorepos
+
+Using project references with monorepo management tools like `lerna` is straightforward - simply add references according to the dependencies of your code. This should generally mirror the dependencies listed in your `package.json`s. Lerna's symlinking will do the rest!
+
+You can see the example repo https://github.com/RyanCavanaugh/learn-a for a cloneable example with a walkthrough file showing how to accomplish various tasks.
```
mhegazy: We should have this in the handbook or on the TypeScriptSamples repo.

PR #790, opened 2018-07-15
The "Classes" section of the "TypeScript in 5 minutes" tutorial minorly corrected

automatix:
The example code in the "Classes" section of the "TypeScript in 5 minutes" tutorial is broken. This commit should repair it.


mhegazy:
Not sure what is "broken" about this sample. the sample is meant to demonstrates the structural comparison between the class type and the interface type.

PR #791, opened 2018-07-16
Improve module templates documentation

### LINKED ISSUE #785: Improve documentation for writing declaration files
The template files for declaration files for modules, specifically `module-class.d.ts` and `module-function.d.ts`, use a syntax as follows:

```
// module-class.d.ts
export = MyClass;
declare class MyClass { /* ... */ }
declare namespace MyClass { /* ... */ }
```

```
// module-function.d.ts
export = MyFunction;
declare function MyFunction(/* ... */): /* ... */;
declare namespace MyFunction { /* ... */ }
```

However, this is specifically called out as incorrect in an ES6-context in the DefinitelyTyped repository's README and in the cited [Stack Overflow answer](https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this) by @RyanCavanaugh (at least as far as I understand it).

What should the correct syntax be? How/where should I define other types or interfaces which are used by the exported class or function?

DanielRose:
Add some documentation on the tradeoff involved when declaring a namespace in a class-module or function-module.
Extend the documentation on correctly importing such a module.

Fixes #785


pages/declaration files/templates/module-function.d.ts.md
```diff
@@ -9,13 +9,18 @@
...
- *~   import x = require('someLibrary');
- *~
- *~ Refer to the documentation to understand common
- *~ workarounds for this limitation of ES6 modules.
- */
+// Note that ES6 modules cannot directly export class objects.
+// This file should be imported using the CommonJS-style:
+//   import x = require('someLibrary');
+//
+// Alternatively, if --allowSyntheticDefaultImports or
```
mhegazy: `--allowSyntheticDefaultImports` is not something we advertise. it is only useful for users targeting `module: system`.

 `--esModuleInterop` is our recommended default. i wounder if we should just change the templates to reflect that.. 

@RyanCavanaugh and @DanielRosenwasser thoughts?

PR #792, opened 2018-07-24
Add documentation for environment variables to configure watching of files and directories

sheetalkamat:
Fixes Microsoft/TypeScript#25715

### DIFF
--- a/pages/Configuring Watch.md
+++ b/pages/Configuring Watch.md
@@ -1,0 +1,30 @@
+## Configuring Watch of file and directory
+
+`fs.watch` and `fs.watchFile` provided by node, both have pros and cons.
+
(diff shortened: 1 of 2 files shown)


pages/Configuring Watch.md
```diff
@@ -0,0 +1,30 @@
+## Configuring Watch of file and directory
```
mhegazy: can you add a link to this page from the compilerOptions page watch section.

pages/Configuring Watch.md
```diff
@@ -0,0 +1,30 @@
...
+
+`fs.watch` and `fs.watchFile` provided by node, both have pros and cons.
```
mhegazy: i would add a heading for this section as `Background`

pages/Configuring Watch.md
mhegazy: do not think you need the header. the page title is sufficient.

pages/Configuring Watch.md
mhegazy: I would start be saying the `--watch` implementation relies on node's underlying fs.watch/fs.watch file.

mhegazy: Also consider moving this section to the bottom.

PR #799, opened 2018-08-01
Update Type Checking Javascript files

sandersn:
Right now this is just copying updates from the wiki, which has had quite a few more changes than the handbook version. I'll be adding the JSDoc tag documentation when I finish documenting class/extends/this/enum.


mhegazy:
Do we still need the pages on the wiki? or can we redirect them to the handbook instead?

sandersn:
I'll update the wiki pages once we re-publish the handbook. @DanielRosenwasser

PR #816, opened 2018-09-04
Revert "Update Compiler Options.md"

NMinhNguyen:
This reverts commit 8e974c814d44381ed05a3b6203894e0f9e39d552.

This supports https://github.com/Microsoft/TypeScript/pull/26866, and thus should not be merged before the code change itself.

I wasn't sure if I'm targeting the right branch with this PR (I couldn't find branch `release-3.1` which I believe @RyanCavanaugh would like the code change to go into), so please let me know if I should change it.

Fixes https://github.com/Microsoft/TypeScript/issues/26193.

/cc @DanielRosenwasser


BurtHarris:
@NMinhNguyen, I'm not sure the documentation is useful either way.    Would it be correct to document -- allowSyntheticDefaultImports as deprecated, and say that --esModuleInterop is a more comprehensive solution to the problem synthetic default imports was trying to address?

PR #825, opened 2018-10-02
Update Type Inference - mouseEvent.button is valid and not a bad example

GuyKh:
Fixes [#26472](https://github.com/Microsoft/TypeScript/issues/26472) (in TypeScript repo)

Updated the `mouseEvent.button` example for Contextual Type inheritance and changed it to `mouseEvent.clickTime` so it would probably not be used in the future


RyanCavanaugh:
Thanks!

PR #831, opened 2018-10-02
Added basic Parcel tutorial

yamalight:
This PR adds basic Parcel tutorial.
I've recently tried getting into TypeScript world and found that Parcel is possibly the simplest way to use it for front-end development.
Currently, there's a tutorial for webpack, but not for Parcel - I though I'd fix that 🤔


RyanCavanaugh:
We can't document every loader on the planet here. This belongs on parcel's pages (https://parceljs.org/typeScript.html)

DanielRosenwasser:
Thanks for the PR though @yamalight - just to clear things up, the problem is that that Webpack guide is already hard to maintain, and nowadays I'd say TypeScript is better off referring to the docs on Webpack, Parcel, etc.

yamalight:
Got it, thanks for feedback!

PR #834, opened 2018-10-03
Change Object to object as it is a non-primitive type

GuyKh:
Fixes [#19908](https://github.com/Microsoft/TypeScript/issues/19908)

Changed the use of Object to object and added the link to Do's and Don'ts


RyanCavanaugh:
This renders the example too weird because the initial assignment becomes an error.

PR #835, opened 2018-10-04
Add styled-components tutorial section

d-ivashchuk:
When I was starting out with typescript and react this guide has really helped me, but I've spent too much time configuring `styled-components` to work properly. This is an attempt to help future learners of typescript+react to move faster if they want to use css-in-js via `styled-components`.


RyanCavanaugh:
We can't document how to use every single library that might interact with React on this page. This is better suited to an off-site blog post or Stack Overflow question.

PR #842, opened 2018-10-05
Fix link

dumblole:
Fixes https://github.com/Microsoft/TypeScript/issues/13177


dumblole:
I did what I did because I believe that the file location has little to do with the actual build website since other files don't belong in their folder path as seen in the built website. 
This was the broken link on the website:https://www.typescriptlang.org/docs/handbook/tsconfig-json.html 
This is what it should be after my fix: https://www.typescriptlang.org/tsconfig-json.html

PR #847, opened 2018-10-05
Avoid unnecessary any casts in advanced types

niedzielski:
- Tighten the typing to avoid unnecessary casts, especially casts to any.
- Use const where possible over mixture of let and var.
- Check Object.hasOwnProperty() in for..in loops.
- Add console.log() invocation so the example prints.

I don't claim the example is perfect but it seems like an improvement.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -10,33 +10,37 @@ You will mostly see intersection types used for mixins and other concepts that d
 Here's a simple example that shows how to create a mixin:
 
 ```ts
-function extend<T, U>(first: T, second: U): T & U {
-    let result = <T & U>{};
-    for (let id in first) {
-        (<any>result)[id] = (<any>first)[id];
-    }
-    for (let id in second) {
-        if (!result.hasOwnProperty(id)) {
-            (<any>result)[id] = (<any>second)[id];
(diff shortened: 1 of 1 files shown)


pages/Advanced Types.md
```diff
@@ -10,33 +10,37 @@ You will mostly see intersection types used for mixins and other concepts that d
...
-    for (let id in second) {
-        if (!result.hasOwnProperty(id)) {
-            (<any>result)[id] = (<any>second)[id];
+    for (const prop in second) {
+        if (second.hasOwnProperty(prop)) {
+            (<Second>result)[prop] = second[prop];
         }
     }
-    return result;
+    return <First & Second>result;
```
SlurpTheo:
```ts
result as First & Second
```

Similar above for:
```ts
result as First
result as Second
```

niedzielski: Is the `as` operator consistent with the rest of the docs' type assertions?

SlurpTheo: It's the more modern way to do type assertions (i.e., works in JSX contexts where the original `<>`-wrapped do not) -- see [`as foo` vs. `<foo>`](https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html).

PR #856, opened 2018-10-09
Removed --stripInternal from documentation

andrewnester:
Fixes https://github.com/Microsoft/TypeScript/issues/27455


PissedCapslock:
@RyanCavanaugh Now that this `--stripInternal` has been removed from the documentation, what is the recommended way to hide certain methods from your public API ?

Assume you write a library in TS, and ship it as JS files with .d.ts definition files.

In the past, you could mark certain **public** methods in your Typescript code base as `@internal`. This allowed to use those methods inside your library, and by removing them from the definition files in your shipped version, users of your library wouldn't see those methods.

In Java, this is typically solved by using package visible methods. As there is no such construct for Typescript, all the solutions you find on the internet refer to this now removed `--stripInternal` option.

Is there an alternative available ?

PS: searching on the `--stripInternal` string still points to the [Compiler Options in MSBuild](http://www.typescriptlang.org/docs/handbook/compiler-options-in-msbuild.html#mappings), so it hasn't been completely removed.

orta:
This CLI option isn't removed, and probably never will be - it's just considered an internal compiler detail. ( see the [new tsconfig docs](https://www.typescriptlang.org/v2/en/tsconfig#stripInternal) ) so it's a buyers beware kind of deal

That said, you might be interested in using https://api-extractor.com/ to do the public/private DTS combo

PissedCapslock:
Thanks for pointing me to the new docs. I like the new styling. And good to know it is still in there.

I'm aware of api-extractor. Alas, due to it's restriction of having a single entry point for your typings, I cannot use it in my project.

PR #895, opened 2018-10-25
Clarify the comment as described in #754

### LINKED ISSUE #754: Clarification of comments in module-class.d.ts
As someone learning about TypeScript declaration files, I was surprised by this comment in the [`module-class.d.ts` template](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/templates/module-class.d.ts.md):

> Note that ES6 modules cannot directly export class objects.

That left me curious about the meaning of the comment, particularly with respect to this type of module:
```js
// SomeModule.js

export default class {
  // etc.
}
```

Is this comment referring to some limitation of ES modules?  Or is it referring to the behavior of TypeScript declaration files?

I'd be happy to help clarify the comments in the templates if someone can help me understand specifically what they mean.

This should probably be a separate issue, but I'm also struck by this syntax in the templates:
```js
// This file should be imported using the CommonJS-style:
import x = require('someLibrary');
```

I opened #753 suggesting a correction to this same syntax elsewhere in the docs.  But seeing it again in the template I'm wondering if it is TypeScript specific syntax (I don't see the same mix of `import` and `require` used on the [module doc](https://www.typescriptlang.org/docs/handbook/modules.html)).

dutchmartin:
Fixes  #754

### DIFF
--- a/pages/declaration files/templates/module-class.d.ts.md
+++ b/pages/declaration files/templates/module-class.d.ts.md
@@ -10,7 +10,7 @@
  */
 
 /*~ Note that ES6 modules cannot directly export class objects.
- *~ This file should be imported using the CommonJS-style:
+ *~ This file should be imported using the TypeScript specific style (instead of CommonJS-style):
  *~   import x = require('someLibrary');
  *~
  *~ Refer to the documentation to understand common


RyanCavanaugh:

PR #896, opened 2018-10-25
Update code samples to compile with stricter rules

tekerson:
A number of the "correct" code samples in the Interfaces document failed to compile with either `noImplicitAny` of `strictPropertyInitialization` enabled.

Fixes https://github.com/Microsoft/TypeScript/issues/28044

### DIFF
--- a/pages/Interfaces.md
+++ b/pages/Interfaces.md
@@ -334,7 +334,7 @@ interface ClockInterface {
 }
 
 class Clock implements ClockInterface {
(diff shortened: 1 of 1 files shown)


tekerson:
In the `ClockInterface` examples I added an initializer to the `currentTime` property to satisfy the `strictPropertyInitialization` check (`currentTime: Date = new Date();`), but maybe it would be simpler to make the property optional (`currentTime?: Date;`)?

PR #901, opened 2018-10-27
note added for --allowSyntheticDefaultImports

### LINKED ISSUE #700: Update docs to reflect `--allowSyntheticDefaultImports`
From https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require:

> When importing a module using export =, TypeScript-specific import module = require("module") must be used to import the module.

The *must* is not always true -- one can use a default import given the right settings.

ajithcpas:
Fixes #700


RyanCavanaugh:
This isn't really accurate enough to just put in as-is.

PR #941, opened 2019-01-04
Fixes 'keyof T' type example

aleph-naught2tog:
Fixes #29249 from the TypeScript repository

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -773,14 +773,14 @@ let unknown = getProperty(person, 'unknown'); // error, 'unknown' is not in 'nam
 ## Index types and string index signatures
 
 `keyof` and `T[K]` interact with string index signatures.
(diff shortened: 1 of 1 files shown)


guillaumep:
@aleph-naught2tog Actually TypeScript returns `(string | number)[]` and not `string[]`.

This atypical result would certainly require some explanation (it confuses me a lot).

aleph-naught2tog:
@DanielRosenwasser I think I'm confused; which is the _correct_ typing then? I had read the issue as being that it _wasn't_ `string[]` so the fix was to make it that, but I can totally change it 😁

RyanCavanaugh:

tommck:
So... it's been 5 months... still not updating the old content?  Do we have an ETA on the new handbook?

orta:
I just merged a PR like this, so the fix will appear in the new site on the next deploy

PR #955, opened 2019-01-26
React function component, not stateless functional

pshrmn:
React no longer refers to function components as "stateless functional components". Functional was changed to function to differentiate from functional programming and "stateless" was dropped with the advent of React Hooks. From the [reactjs.org documentation](https://reactjs.org/docs/hooks-state.html#hooks-and-function-components):

> You might have previously known these as “stateless components”. We’re now introducing the ability to use React state from these, so we prefer the name “function components”.

This PR replaces all instances of "stateless functional component" and "sfc" with "function component".

### DIFF
--- a/pages/JSX.md
+++ b/pages/JSX.md
@@ -104,12 +104,12 @@ import MyComponent from "./myComponent";
 
 There are two ways to define a value-based element:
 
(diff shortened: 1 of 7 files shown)


RyanCavanaugh:

PR #965, opened 2019-02-12
Amend Overloads in Functions chapter

### LINKED ISSUE #964: Overload signatures
This information that I found in [this blog](https://www.sitepen.com/blog/update-the-definitive-typescript-guide/) is missing in the handbook:

> It is extremely important to keep in mind that the concrete function implementation must have an interface that matches the lowest common denominator of all of the overload signatures.

I'd like to add this to the Handbook, since it led to lots of unnecessary error debugging. The example is working, but it's not detailed why.

The error is "Overload signature is not compatible with function implementation." - where it's not trivial that the last `function` line (plus the function body) is considered "implementation".

gombosg:
Add that the function implementation must be a common denominator of all overloads. 
This is an important fact that can lead to compiler errors.

Fixes #964

### DIFF
--- a/pages/Functions.md
+++ b/pages/Functions.md
@@ -466,7 +466,10 @@ With this change, the overloads now give us type-checked calls to the `pickCard`
 In order for the compiler to pick the correct typecheck, it follows a similar process to the underlying JavaScript.
 It looks at the overload list, and proceeding with the first overload attempts to call the function with the provided parameters.
 If it finds a match, it picks this overload as the correct overload.
+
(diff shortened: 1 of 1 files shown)


RyanCavanaugh:
This isn't worded clearly enough to improve the page. Can you try to rephrase? I don't think the edit in the top section is the right place for the added section.

orta:
It's been four months, I'm happy to call this stalled - @gombosg - thanks for contibuting, we're open to you taking  second shot at this still.

PR #969, opened 2019-02-17
Fix unencoded spaces in links

### LINKED ISSUE #533: All links included spaces are broken
Probably because of the breaking changes of gfm.

e.g. https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Compiler%20Options.md

mezzode:
Multiple links were broken on GitHub since GitHub Flavored Markdown needs spaces in links to be encoded. This is an updated version of #898.

It might also be useful to add a [custom lint rule](https://github.com/DavidAnson/markdownlint/blob/master/doc/CustomRules.md) checking for this, but that deserves its own issue/pr.

Fixes #533.


orta:
Thanks!

Issue #377, opened 2016-09-01
As a reader, I would like this handbook in epub format


connerbw:
Can you make this book available in Epub (or mobi) format?

DanielRosenwasser:
Sorry for the delay. I've definitely wanted this in the past. You can definitely do this on your own using Pandoc or Gitbook. I think we'd accept PRs on this as well.

vgvenkat:
Can I work on this if its not already taken? I am trying to do gitbook, will update if it goes well!

DanielRosenwasser:
Just saw this, but sure, go for it!

olavolav:
Just wrote a Ruby script to do this using [pandoc](https://pandoc.org/epub.html), see below. Let me know if you'd like to have this as a PR in some way.

```ruby
# Define chapters

chapters = [
  "Basic Types",
  "Variable Declarations",
  "Interfaces",
  "Classes",
  "Functions",
  "Generics",
  "Enums",
  "Type Inference",
  "Type Compatibility",
  "Advanced Types",
  "Symbols",
  "Iterators and Generators",
  "Modules",
  "Namespaces",
  "Namespaces and Modules",
  "Module Resolution",
  "Declaration Merging",
  "JSX",
  "Decorators",
  "Mixins",
  "Triple-Slash Directives",
  "Type Checking JavaScript Files"
]


# Reformat Markdown sources

full_content = <<-HEREDOC
---
title: TypeScript Handbook
author: Microsoft Corporation
rights: Apache License 2.0
language: en-US
---
HEREDOC

chapters.each do |chapter|
  file_name = "#{chapter}.md"
  content = File.read(file_name)

  # Increase section level of headings
  content = content.gsub(/(#){1,5}\s([^\n]*)/, '#\1 \2')

  # Add top-level heading
  content = content.prepend("# #{chapter}\n\n")

  full_content << content
end

File.write("all.md", full_content)

# Generate epub file

markdown_file_names = chapters.map { |chapter| "'#{chapter}.md'" }.join(" ")

system("pandoc -s -o ~/Desktop/TypeScript-Handbook.epub all.md")

File.delete("all.md")
```

olavolav:
@DanielRosenwasser Could you clarify what kind of PRs you'd welcome on this issue? I suppose for example one could rewrite the Ruby script as Typescript, and then perhaps move this to a new `scripts` folder.

What do you think?

hanspoo:
[...]

PR #972, opened 2019-02-23
[WIP] Generate an ebook using GitBook

salomvary:
Based on the draft solution [proposed here](https://github.com/Microsoft/TypeScript-Handbook/issues/377#issuecomment-443398371) I managed to generate an ebook from the Handbook using the following steps:

- Install `ebook-convert` ([see instructions here](https://toolchain.gitbook.com/ebook.html))
- `npm install`
- `npx gitbook epub . typescript-handbook.epub`

I had to add a title to each page, otherwise there is nothing in the ebook showing the beginning of a chapter. This will probably introduce a problem with duplicate titles when the pages are used for generating typescriptlang.org. (The problem actually [already exists on certain pages](https://www.typescriptlang.org/docs/handbook/variable-declarations.html)). It would be nice to have the chapter title on each page anyway in my opinion.

The main page titles and internal headings would benefit from a general cleanup as they are not consistent. I would recommend using `<h1>` for the chapter title only, and use `<h2>` or deeper for subtitles.

TODO

- [x] Add metadata to the book (https://toolchain.gitbook.com/config.html)
- [ ] Add cover page (https://toolchain.gitbook.com/ebook.html)
- [ ] Apply consistent headers in all pages
- [ ] Check if all chapters are included in the SUMMARY.md (maybe automatically?)
- [ ] "Proofread" the ebook, look for errors
- [ ] Automatically publish the ebook on typescriptlang.org (?)

Here is the work-in-progress ebook: [typescript-handbook.epub.zip](https://github.com/Microsoft/TypeScript-Handbook/files/2896773/typescript-handbook.epub.zip)


orta:
At this point, it's probably not worth it - given that there'll be [a new handbook](https://github.com/microsoft/TypeScript-new-Handbook), but I think this sorta infra would be useful for the new handbook once it's sent up

Thanks for giving it a shot, I think we should have an offline book people can have for a reference

PR #974, opened 2019-02-25
Make TS example equivalent to JS example

jackwilsdon:
There is no open issue for this, but I noticed that the TypeScript example is not equivalent to the JS above it;

```JavaScript
let pet = getSmallPet();

if (pet.swim) {
    pet.swim();
}
else if (pet.fly) {
    pet.fly();
}
```


orta:
Agree, thanks!

PR #979, opened 2019-02-28
Quotes used consistently in gulp tutorial.

flatiron32:
While working through the gulp tutorial I noticed that the `gulpfile.js` `require` instructions were not all using the same quoting. This moves all the TS and JS to use single-quotes.


orta:
Cool, thanks!

PR #980, opened 2019-03-02
Changed explanation for 'reference types' directive

marckassay:
Changed explanation about `/// <reference types="..." />` to read a little better.


orta:
Cool, thanks!

PR #984, opened 2019-03-05
React tutorial follows interface naming lint rules that require 'I' prefix

flatiron32:
The lint rules expect I to prefix interfaces.


orta:
Hi @flatiron32  - I think overall we're fine without these changes, it'd have to be consistent all over the docs and (personally) I don't think it adds much in a structural type system where almost everything is an interface.  So I'm going to close this PR, but thanks for helping to improve out docs.

PR #986, opened 2019-03-11
Use `textContent` instead of `innerHTML` in the 5-min example.

lgarron:
Using `innerHTML` to write to the document is dangerous practice in general, and
it would be best to avoid giving beginners the impression that they should use
it directly.

In these examples, `textContent` is safer, and more clearly communicates the
the intent to write a string to the DOM.


orta:
Yeah, I think this is nicer - thanks

PR #988, opened 2019-03-14
Replace deprecated type SFC with FunctionComponent

a-xin:
React's Function Components cannot be considered stateless anymore, therefore the type SFC has been deprecated.

This PR replaces references to `SFC` with `FunctionComponent` accordingly.

https://reactjs.org/docs/hooks-state.html#hooks-and-function-components


orta:
Looking good, thanks for the comprehensive PR!

PR #989, opened 2019-03-17
Fix union type description in Tuple part of Basic Type chapter

xiaojingzhao:
Fixes [#28894](https://github.com/Microsoft/TypeScript/issues/28894)


orta:
We now mention that it fails with an error instead - thanks for helping to improve the dos though @xiaojingzhao

PR #995, opened 2019-03-21
Fix grammatical issue

SCKelemen:
will get run -> will run


SCKelemen:
Also, why doesn't exhaustiveness work here?
Screen Shot 2019-03-21 at 2 55 39 PM

orta:
Unusure about the switch exhaustiveness - but thanks for the PR!

PR #998, opened 2019-03-28
Fix tuple definition

### LINKED ISSUE #975: Is Tuple union type still work?
On http://www.typescriptlang.org/docs/handbook/basic-types.html , It said
> When accessing an element outside the set of known indices, a union type is used instead

However, when I tested this example locally, both `tsc` and `deno` threw an error:

> Type '"some string here"' is not assignable to type 'undefined'

Here is the example:
```
let x: [string, number];
x = ["hello", 10];
console.log('x[0].substr(1)', x[0].substr(1));
x[2] = 'world'; // Error was threw here
```

hateonion:
## Motivation
Per this comment https://github.com/Microsoft/TypeScript/issues/28894#issuecomment-445370682 , Tuple type should be fixed width in [Typescript 2.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html). Which means the current document for Tuple type is outdated, My update is to fix this.

## What I really did in this pull request
1. Modified the code samples for the Tuple type, add more comments, trying to give a more detailed declaration for Tuple Type. e.g. The first time I read the doc, I'm quite confused how Tuple type actually works. I was thinking it's actually used for declaring some key-value pair, but actually, it's for declaring an array which each of the element has a specific type.
2. As mentioned above, the Tuple definition is outdated, now access the index outside of the defined one in a Tuple is not allowed anymore, and this will resolve #975 and another issue in Typescript repo https://github.com/Microsoft/TypeScript/issues/28894#issue-388536852

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -92,17 +92,13 @@ console.log(x[0].substr(1)); // OK
 console.log(x[1].substr(1)); // Error, 'number' type does not have 'substr' method
 ```
 
-When accessing an element outside the set of known indices, a union type is used instead:
+Around Typescript 2.5, Tuple became fixed length. Accessing an element outside the set of known indices is not allowed anymore :
 
 ```ts
-x[3] = "world"; // OK, 'string' can be assigned to 'string | number'
-
-console.log(x[5].toString()); // OK, 'string' and 'number' both have 'toString'
-
-x[6] = true; // Error, 'boolean' isn't 'string | number'
+x[2] = "world"; // Error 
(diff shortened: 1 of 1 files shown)


nbkhope:
There's a typo here https://github.com/microsoft/TypeScript-Handbook/pull/998/files#diff-b482036f3eac612448112805b2d12eaeR84

orta:
Thanks! I've merged a PR which covers this https://github.com/microsoft/TypeScript-Handbook/pull/1030 - so this can be closed <3

PR #1001, opened 2019-03-29
Add the recommendation not to use the Symbol wrapper object type

vkrol:
Add the recommendation not to use the Symbol wrapper object type.


orta:
Thanks 👍

PR #1003, opened 2019-04-03
Clarify module loaders in Modules.md's Introduction

CraigMacomber:
Clarify that web apps can use AMD modules with RequireJS. Previously one could read this implying that CommonJS was commonly used with both Node.js and with RequireJS.


orta:
Cool, thanks!

PR #1004, opened 2019-04-04
Adds an example for a special string use case (using an email address).

gewoonwoutje:
I was looking for this use-case for a while and in the end constructed it myself like below. It adds data validation as a type to a dynamic string.

It may be kind-of-a-hack, and I don't know if this is the right place for it. I hope putting it here helps others who are looking for something similar.


orta:
Hi @gewoonwoutje - thanks for helping to improve the handbook! I had a chat with the team and we do want to figure out nominal typing like this example uses, but probably don't want to be encouraging it too much in the handbook until there's something a bit more solid in the language level support. So, I'm going to close this PR.

PR #1006, opened 2019-04-06
fix: Don’t generate table in `readonly` tuples description

ExE-Boss:
Previously, the 3.4 release notes **`readonly` tuples** section would contain a `<table>`: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#readonly-tuples

This fixes that and also improves the line’s formatting.

---

review?(@sandersn)


orta:
Reads bette rto me in the markdown preview, so I'm gonna take this as a +1

PR #1008, opened 2019-04-06
Fix typo

kbsanders:
For consistent grammar with the "Targeting ECMAScript 2015 and higher" section that follows.


orta:
Agree, thanks

PR #1009, opened 2019-04-08
Fix typo - Missing backtick in template string

rmi22186:
There is a missing backtick in the template string in the docs for Basic Types.md. This PR resolves that.


mtreinik:
The beginning backtick is not missing. It is on line 44 of the file, as the example is a multi-line template string.

orta:
Agree, thanks @mtreinik - and thanks @rmi22186 for trying to improve our docs

PR #1010, opened 2019-04-08
Fix error message when passing an object literal

ahmedkrmn:
In the [Excess Property Checks](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks) section in the Interfaces page, when an object literal is passed with a wrong key, the error message currently in the documentation isn't as verbose as the one actually displayed by TypeScript.

This PR fixes the error message.


orta:
Cool, yep

PR #1013, opened 2019-04-14
Update Declaration Merging.md

shawnkoon:
An example provided for Module Augmentation was not correct, updated javascript code to be javascript, then update the typescript example accordingly as well.


orta:
Thanks!

PR #1016, opened 2019-04-19
fixed incorrect explanation of tuple indexing when index is unknown

jaked:
Fixes #


jaked:
oops I didn't realize that pushing to this branch would add to this PR. let me make separate branches.

PR #1017, opened 2019-04-19
fixed incorrect explanation of tuple indexing when index is unknown

jaked:
Fixes #


orta:
I've just merged a PR which also addresses this in a similar way, so I think this is fine to be closed - thanks @jaked !

PR #1018, opened 2019-04-19
explain tuple destructuring

jaked:
Fixes #


orta:
Yeah, this makes sense to me 👍

PR #1024, opened 2019-04-28
replace hard-coded secret worst-practice in class accessor example

sethveale:
Fixes #
The accessor example for Classes.md in the handbook was checking an ambient plain-text password against a hard-coded literal in a property accessor. I just changed it to a simple string length validation so that there's no risk that there's no implied conflation of authentication and authorization, or of authorization implementation that'd never exist in a professional context. Have confirmed that it compiles and runs, and checked the contribution guidelines.

Thanks for the great docs! Sorry for there not being an issue for this - seemed like a little fix.

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -302,14 +302,14 @@ if (employee.fullName) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


orta:
Hah, yeah, this seems like a good call to me

sethveale:
:)

PR #1028, opened 2019-05-02
Clarify intent of --mapRoot

Rolias:
Add two sentences to make it clear that `--mapRoot` does not create the map files in the specified path.

Fixes #1207


orta:
This feels reasonable to me, somehow I can't get to the original issue  to verify but I'm open to this being the right behavior.

PR #1029, opened 2019-05-02
"Variable Declarations.md" uses "?" before it is defined in "Functions.md"

colohan:
"Variable Declarations.md" uses "?" before it is defined in "Functions.md".  Adding a brief explanation so readers new to Typescript can read this documentation in-order without getting confused (as I did).

Fixes #


orta:
Yeah, that works for me. Thanks @colohan

PR #1030, opened 2019-05-07
Update tuple docs to match post-2.7 behaviour

IvanVergiliev:
Tuples were changed in 2.7 (see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html , section "Fixed Length Tuples") to not allow assignments outside of the declared indexes. This commit updates the handbook docs to match this behaviour.

Fixes #


orta:
Looking good - thanks 👍 

Verified all this in the playground

PR #1031, opened 2019-05-08
fixed two small errors

0xflotus:
Fixes no known issue but two small errors.


orta:
Yep, looks good - thanks

PR #1043, opened 2019-05-30
[doc] typo in Declaration Merging

Oliboy50:
Updating [https://www.typescriptlang.org/docs/handbook/declaration-merging.html](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)

`observable.js` should be `observable.ts` since it includes TypeScript code

```ts
export class Observable<T>
```

and the file is referenced below with `// observable.ts stays the same` which make it clear that we expect the file to end with `.ts`


orta:
Yeah, I agree - thanks

PR #1045, opened 2019-06-02
Add emphasis to release notes

cinderblock:
This page is at the top of many search results. This "highly recommended" suggestion is buried on the page. Can we make it more prominent?

-----

PS. Sorry for jumping straight to a PR but this is so simple I thought it was not worth opening an issue too.

PPS. I noticed the suggestion of keeping sentences alone on single lines in the guide but I also wanted to change as little as possible...


orta:
Cool, yeah, works for me - thanks

PR #1056, opened 2019-06-27
Make the React & Webpack tutorial cross-platform compatible

mbertram:
mkdir doesn't support -p on Windows. This is a cross-platform way of creating the folder structure.


orta:
👍 thanks

PR #1057, opened 2019-07-01
update interface vs type aliases docs, to include features supported …

cryptoB0T:
…since ts 2.7

Fixes #
The docs should indicate that type aliases can be extended, as this is supported since 2.7.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -483,7 +483,8 @@ declare function aliased(arg: Alias): Alias;
 declare function interfaced(arg: Interface): Interface;
 ```
 
-A second more important difference is that type aliases cannot be extended or implemented from (nor can they extend/implement other types).
+Type aliases couldn't be extended or implemented from (nor could they extend/implement other types), in older versions of TypeScript. As of version 2.7, type aliases can be extended. Teams developing React applications sometimes choose to use `Type` aliases over `Interface`.
+
 Because [an ideal property of software is being open to extension](https://en.wikipedia.org/wiki/Open/closed_principle), you should always use an interface over a type alias if possible.
 
 On the other hand, if you can't express some shape with an interface and you need to use a union or tuple type, type aliases are usually the way to go.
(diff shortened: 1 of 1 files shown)


orta:
I tried looking through the release notes but I couldn't find anything about this feature, do you mean something [like this](https://www.typescriptlang.org/play/#code/C4TwDgpgBAsiAq5oF4oG8BQUoEMBcUARgPbEA2EOAdhgL4YaiSwgCiAHsBFQCYQ+JmqOIOgAydEQIBnYACcAllQDmUehj4BjMjjnRNxKrKgRO3PjwJwOXXv1ENTtiwDocGJ+f4vCGIA)?

```ts
type MyType = {
  a: boolean
}

type MyExtendedType = MyType & { b: string }

declare const extended: MyExtendedType

extended.a
extended.b
```

cryptoB0T:
yes, this is correct. the docs appear to indicate that type aliases cannot be extended. it doesn't appear in the release notes but the TS compiler seems to support this functionality. is there something I'm missing?

pages/Advanced Types.md
```diff
@@ -483,7 +483,8 @@ declare function aliased(arg: Alias): Alias;
 declare function interfaced(arg: Interface): Interface;
 ```
 
-A second more important difference is that type aliases cannot be extended or implemented from (nor can they extend/implement other types).
+Type aliases couldn't be extended or implemented from (nor could they extend/implement other types), in older versions of TypeScript. As of version 2.7, type aliases can be extended. Teams developing React applications sometimes choose to use `Type` aliases over `Interface`.
```
orta:
```suggestion
in older versions of TypeScript, type aliases couldn't be extended or implemented from (nor could they extend/implement other types). As of version 2.7, type aliases can be extended by creating a new intersection type e.g. `type Cat = Animal & { purrs: true }`.
```

orta:
Yeah, I also was a little stumped - but that's fine, I've tried to update the PR to include that info with an example to make it more concrete 👍

orta:
Thanks!

PR #1058, opened 2019-07-02
fix the non-enumerable of the class prototype

a417420427:
Fixes #


a417420427:
it's related to [https://github.com/microsoft/TypeScript/issues/32183](url)

PR #1061, opened 2019-07-06
Resolve large font size in handbook documentation

justinpage:
The following address was showing a large font for import resolutions:

https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

Instead of using an ordered list within a bullet point, bring the
subject line to the parent level and then use an ordered list.


orta:
Great, thanks for the fix! I remember seeing this once and not being able to figure out why

PR #1065, opened 2019-07-08
Add `Omit<T,K>` which was introduced in TypeScript 3.5

### LINKED ISSUE #1064: Missing Omit<T,K> in utility types documentation
`Omit<T,K>` [was introduced in TypeScript 3.5](https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#the-omit-helper-type) but is missing in the documentaion.

arthow4n:
Fixes #1064

### DIFF
--- a/pages/Utility Types.md
+++ b/pages/Utility Types.md
@@ -8,6 +8,7 @@ TypeScript provides several utility types to facilitate common type transformati
 * [`Readonly<T>`](#readonlyt)
 * [`Record<K,T>`](#recordkt)
 * [`Pick<T,K>`](#picktk)
+* [`Omit<T,K>`](#omittk)
 * [`Exclude<T,U>`](#excludetu)
 * [`Extract<T,U>`](#extracttu)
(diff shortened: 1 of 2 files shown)


pages/Utility Types.md
```diff
@@ -109,6 +110,27 @@ const todo: TodoPreview = {
 };
 ```
 
+# `Omit<T,K>`
+
+Constructs a type by picking all properties but `K` from `T`.
```
orta:
```suggestion
Constructs a type by picking all properties from `T` and then removing `K`.
```

Mainly so that the order of `T` and `K` are the same, I had to re-read twice to get it with them opposite

arthow4n: I agree with you, thanks for the feedback!

arthow4n: I agree with you, thanks for the feedback! Have amended this to the commit.

orta:
Thanks!

PR #1071, opened 2019-07-16
fix: typo in jsdoc @template example

sbdchd:
The param name is incorrect.

Currently says `p1` instead of the function's param `x`


orta:
Thanks!

PR #1072, opened 2019-07-17
Update Modules.md

willyngh:
1 - rename 'Validation.ts' to 'StringValidator.ts' for better understanding of 'AllValidators.ts'.
2 - import {StringValidator} to correct the error 'Can not find name 'StringValidator' in the module 'ZipCodeValidator.ts'.
3 - changes in the module 'AllValidators.ts' for correct reference and expansion of comments of the respective exports

Fixes #


orta:
Agree -thanks!

PR #1073, opened 2019-07-17
Fix Index types and string index signatures section

guillaumep:
The documentation is incorrect. See https://github.com/microsoft/TypeScript/issues/29249

Fixes #29249 from the TypeScript repository.


orta:
Cool - yep confirmed. 

Thanks!

PR #1079, opened 2019-07-24
Complete the index signature doc.

gxvv:
Related: microsoft/TypeScript#29249

That issue has corrected the error doc of string index signature. But it isn't mentioned that the number index signature has different result from string index signature. This pr append this situation.


orta:
Thanks @gxvv great work  - confirmed on the playground 👍

PR #1080, opened 2019-07-24
[Compiler Options] Fix dead link to tsconfig.json

### LINKED ISSUE #1077: Broken link in Compiler Options html
At the bottom of the [Compiler Options page](https://www.typescriptlang.org/docs/handbook/compiler-options.html), the tsconfig.json link leads to a 404 with https://www.typescriptlang.org/docs/handbook/tutorials/tsconfig-json.html.

The right link is https://www.typescriptlang.org/docs/handbook/tsconfig-json.html.

Since the tsconfig.json.md source is in the tutorials folder, I don't know if it's the link that is broken or if the source must be published.

neilbryson:
Fixes #1077 

Tested change in my local TypeScript-Website repo

Downside is navigation through Markdown will be broken.

Will this be a desirable fix for this ?

### DIFF
--- a/pages/Compiler Options.md
+++ b/pages/Compiler Options.md
@@ -98,5 +98,5 @@ Option                                         | Type      | Default            
 
 ## Related
 
-* Setting compiler options in [`tsconfig.json`](./tutorials/tsconfig.json.md) files.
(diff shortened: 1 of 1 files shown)


orta:
I think more people read the site than the markdown - let's do it. Thanks!

PR #1083, opened 2019-07-29
remove unnecessary space from modules handbook document

miki725:
in backticks string, there is already space defined after `"{s}"` so there no need for additional space in the subsequent conditional expression

previous behavior:

```js
> function validate(s) { return s.length === 5 }
undefined
> ["Hello", "98052", "101"].forEach(s => { console.log(`"${s}" ${validate(s) ? " matches" : " does not match"}`); })
"Hello"  matches
"98052"  matches
"101"  does not match
undefined
```


orta:
Thanks!

PR #1087, opened 2019-08-03
Remove a comment about union types in tuple doc

FujiHaruka:
As a result of #1030, the comment about union type in tuple doc is no longer necessary.


orta:
Cool, thanks!

PR #1093, opened 2019-08-12
Fix broken layout of Table of Contents

zaki-yama:
Fixes broken layout of ToC in some pages.

Example in [JSX](https://www.typescriptlang.org/docs/handbook/jsx.html) page.
SS_2019-08-13_3_17_42


orta:
Thanks!

PR #1094, opened 2019-08-14
Add code sample and clarify some expression

srcdes:
The sentence `the type checker would have warned us` might mislead about "compiler warning". To clarify its correct meaning, i also have added an example related to it.


orta:
Thanks, let's do this 👍

PR #1053, opened 2019-06-24
Fix documentation on --strictNullChecks

karlhorky:
Ref (documentation): https://www.typescriptlang.org/docs/handbook/compiler-options.html
Ref (issue in TypeScript repo): https://github.com/microsoft/TypeScript/issues/16075#issuecomment-304763298

cc @RyanCavanaugh 

Closes microsoft/TypeScript#16075
Closes microsoft/TypeScript#11758

PR #1096, opened 2019-08-20
Emphasize that void accepts null unless strictNullChecks

wataash:
Though this specification was written in #1053, I still got confused when I read the section "Void" -- "`null` can't be assigned to `void` on TypeScript playground... is document wrong?"
And I had stayed confused for 10 minutes until I carefully read the next section (Null and Undefined).

This changes remove the time window of confusing between these two sections, complementing #1053.

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -179,10 +179,11 @@ function warnUser(): void {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


orta:
👍 works for me, thanks @wataash

PR #1110, opened 2019-09-18
Update docs for --isolatedModules flag

### LINKED ISSUE #1109: --isolatedModules docs are incorrect
The [compile options docs](https://www.typescriptlang.org/docs/handbook/compiler-options.html) for `--transpileModule` seem to be incorrect.

They say 'Transpile each file as a separate module (similar to “ts.transpileModule”)' but that is not what the flag does. Instead, it does extra checking to ensure that separate compilation would be safe.

Opened PR #1110 to update the docs.

mheiber:
Make it clear that --isolatedModules does *not*
do separate compilation, but instead ensures
that separate compilation would be safe.

Also include links to ts.transpileModule
and babel plugin-transform-typescript.

Fixes #1109


orta:
Cool, yeah, this makes sense to me - thanks max

Issue #1039, opened 2019-05-15
Add table of contents for each page


golopot:
A TOC of section headings would help readers greatly.

Swap76:
Hey, I am newbie will love to try this out. Just need a bit info on this :-)

PR #1114, opened 2019-10-03
Added Table of Contents in Basic Types

Swap76:
Added Table of content in Basic Types.md

If this is what is perfect will do the same in all other files else will make as per your request.

This addresses issue #1039


orta:
Yeah, I think this makes sense for now - though in the future we should really have this be automated and as a sidebar

PR #1115, opened 2019-10-03
fix: update `let` section

jsjoeio:
The section on `let` feels outdated. 

> The `let` keyword is actually a newer JavaScript construct that TypeScript makes available.

I rewrote it based on my best solution but hoping to get feedback :)

Fixes #


orta:
Yeah, this work for me. 👍

PR #1117, opened 2019-10-06
Fix typo in JSX Class Component

merapi:
`FactoryFunction` should be either `MyFactoryFunction` to match the previously defined function or just generic `factory function`


orta:
Yeah, good point - thanks!

PR #1122, opened 2019-10-16
fix: implementation missing interface return type

gbili:
the interface is defined as returning boolean, yet implementation is lacking the return type.

Fixes #


orta:
Thanks @gbili  - the return type is inferred in this case, and doesn't need to be included

Issue #1025, opened 2019-04-30
Header links in basic types are broken


dragomirtitian:
Header links on the basic types page are broken, they link to nonexistent pages instead of sections in the current page. 

Ex: The link on Boolean is: `'Boolean'`. Leading to https://www.typescriptlang.org/docs/handbook/Boolean.

From what I can see #838 added these links but they probably should point to ids in the page (#Boolean not Boolean).

dragomirtitian:
I would like to submit a PR for this, but I would like to be able to test it properly. Since linking to sections would use generated ids I would like to test with the generated HTML but I am not sure what tool you use to convert markdown to html.

PR #1123, opened 2019-10-16
Fix Basic Types' table of contents and adds object usage note

ndresx:
This fixes the [Basic Types](https://www.typescriptlang.org/docs/handbook/basic-types.html)' table of contents _(randomly stumbled upon this while fixing the original issue)_.

Moreover, it adds a note to the Any section to avoid using `Object` in favor of `object`. I think it's fine to leave the example as is since you can obviously use it like this, but you shouldn't.

Fixes https://github.com/microsoft/TypeScript-Handbook/issues/1025, https://github.com/microsoft/TypeScript/issues/19908


jadengore:
Good catch @ndresx! Just noticed this myself in the docs and was going to open a PR, just beat me to it 🙂

PR #1126, opened 2019-10-21
Add `bigint` to primitive list

dosentmatter:
`bigint` is a primitive since is not assignable to `object`. Not sure if we want this here though since there is no description for newer types such as `bigint` or `unknown` in "Basic Types.md". [`bigint` is introduced in TS@3.2 page](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html).

```ts
const foo: object = BigInt(100); // Type 'bigint' is not assignable to type 'object'.ts(2322)
```

https://mariusschulz.com/blog/the-object-type-in-typescript

### DIFF
--- a/pages/Basic Types.md
+++ b/pages/Basic Types.md
@@ -239,7 +239,7 @@ function infiniteLoop(): never {
 
 # [Object](Object)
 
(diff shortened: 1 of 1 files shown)


orta:
Agree, thanks!

PR #1132, opened 2019-11-05
Expanded structure docs for project references

JakeGinnivan:
* Make it clear files should be empty rather than indirectly applied

Based on answer in https://github.com/microsoft/TypeScript/issues/34889


orta:
👍

PR #1134, opened 2019-11-05
Add useDefineForClassFields to Compiler Options

sandersn:
I didn't change the MSBuild page because as far as I know MSBuild doesn't support the new flag yet.

Edit: Now it does, so I added it.


sandersn:
I'm going to merge this. Let me know if you have additional comments.

PR #1139, opened 2019-11-06
Fix the playground link under Recursive Type Aliases section

viksorensen:
This fixes the playground link under the Recursive Type Aliases section in the Typescript 3.7 Release Notes


viksorensen:
Duplicate. Removing

PR #1256, opened 2019-11-16
Renaming `function f<T>` to `function foo<T>`

JulianG:
I was reading this page and I believe the function name in the declaration should be `foo` to match the implementation and the following paragraph.

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -1084,7 +1084,7 @@ interface Foo {
     propB: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


orta:
I don't think it is meant to (see how `f()` gets called later) - though having the two functions. so closely names is an anti-pattern, I'd totally take a PR which changes the name of that declare and internal function call though!

JulianG:
oh! absolutely! I totally missed that! sorry!

PR #1257, opened 2019-11-18
Add Parameters<T> to Utility Types

Gerrit0:
Exists since at least 3.1.6, but appears to have been missed.

No related issue, I just noticed it was missing.


orta:
Great catch 👍

PR #1261, opened 2019-11-29
tsconfig.json: extends path allows Node resolution

sirreal:
Add a phrase indicating that the `extends` option may use Node.js tyle module resolution.

This was surprising to me based on the current language in the handbook, but observed behavior and source seem to indicate that Node.js module resolution is allowed:

https://github.com/microsoft/TypeScript/blob/7c14aff09383f3814d7aae1406b5b2707b72b479/src/compiler/commandLineParser.ts#L2402-L2406


orta:
👍 cool, thanks

PR #1266, opened 2019-12-10
Updated Readme.md

vivekverma007:
Added additional badges to respected markdown file.


orta:
Thanks for the PR - but I don't even think the badge we have on here makes sense, so adding more shouldn't really happen.

PR #1270, opened 2019-12-24
Fix Parameter Utility documented return types

ArnaudWeyts:
The return types commented next the `Parameter` Utility type code snippet had 2 typo's in it.


orta:
Cool, thanks - yeah, this makes sense!

PR #1272, opened 2019-12-25
Add article for consistency and fix TypeScript capitalization

### LINKED ISSUE #1271: Small English inconsistencies in Variable Declarations
The line `That’s not to say that block-scoped variable can never be declared with a function-scoped variable.` probably needs an article before `block-scoped variable`. For example, `a block-scoped variable`.

In the last line, `Second, the Typescript compiler...`, Typescript should be TypeScript.

johnjago:
Fixes #1271

### DIFF
--- a/pages/Variable Declarations.md
+++ b/pages/Variable Declarations.md
@@ -290,7 +290,7 @@ function g() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


orta:
Ace, thanks!

PR #1279, opened 2020-01-10
fix Class doc: update Greeter to greeter

bmeritter:
Fixes #


orta:
I think this is the same issue as  #1280 in that it is about the original class function, which means this change doesn't need to occur.

PR #1280, opened 2020-01-10
fix Class doc: update Greeter to greeter

bmeritter:
Fixes #

### DIFF
--- a/pages/Classes.md
+++ b/pages/Classes.md
@@ -472,7 +472,7 @@ greeter = new Greeter("world");
 console.log(greeter.greet());
 ```
 
-Here, `let Greeter` is going to be assigned the constructor function.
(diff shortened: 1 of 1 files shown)


orta:
I think (though it does feel ambiguous) that this sentence refers to the first line in the code sample which means this fix is probably wrong. So I'll close this - thanks for giving an improvement a shot! It means a lot when folks help improve the core docs :D

bmeritter:
@orta  I just think Greeter is a Class, but greeter is an object, greeter is an instance of Greeter. We should let it clear.

orta:
Sure if you want to rename the instance in the docs that could help?

bmeritter:
@orta Yeah

PR #1287, opened 2020-01-28
add results to example code

toshihidetagami:
Those 3 code examples are a bit hard to follow. So I think showing console results helps,
especially for people who are new to OO language.


orta:
Yeah, agreed - thanks!

PR #1288, opened 2020-02-01
Remove references to changes from TypeScript 1.5 in the modules docs

orta:
Does some minor spring cleaning after I spotted this Reddit thread: https://www.reddit.com/r/typescript/comments/ewjhiy/typescript_i_love_you_but_this_documentation_is/

- I tried to orient folks to understanding that namespaces are a TS thing
- I re-oriented the order in which they're presented to give modules first
- I removed the references to the 1.5 transition

### DIFF
--- a/pages/Namespaces.md
+++ b/pages/Namespaces.md
@@ -16,18 +16,10 @@
 [Working with Other JavaScript Libraries](#working-with-other-javascript-libraries)
 * [Ambient Namespaces](#ambient-namespaces)
 
-> **A note about terminology:**
-It's important to note that in TypeScript 1.5, the nomenclature has changed.
-"Internal modules" are now "namespaces".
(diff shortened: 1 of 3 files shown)


pages/Namespaces.md
```diff
@@ -87,7 +77,6 @@ Conversely, the variables `lettersRegexp` and `numberRegexp` are implementation
 In the test code at the bottom of the file, we now need to qualify the names of the types when used outside the namespace, e.g. `Validation.LettersOnlyValidator`.
 
 ## Namespaced Validators
-<b><a href="#table-of-contents">↥ back to top</a></b>
```
orta: These are every 10-20 sentences, makes them pretty distracting. Won't be a problem in v2 handbook either way

PR #1298, opened 2020-02-22
Fix typo - add missing backtick

yeonjuan:
It seems there is a missing backtick(`) character.


orta:
Thanks! this is fixed in the v2 site

PR #1303, opened 2020-03-03
Fix code sample in Metadata section of Decorators page.

### LINKED ISSUE #1302: Samples code in "Metadata" section of "Decorators" page is wrong
Currently (with comments added to explain the issue):

```ts
function validate<T>(target: any, propertyKey: string, descriptor: TypedPropertyDescriptor<T>) {
    let set = descriptor.set;
    descriptor.set = function (value: T) {
        let type = Reflect.getMetadata("design:type", target, propertyKey); // "target" should be "this"
        if (!(value instanceof type)) {
            throw new TypeError("Invalid type.");
        }
        set.call(target, value); // "target" should be "this"
    }
}
```

We should use `this` instead of `target` in those two lines with comments. I will make a PR for it later.

JeffreyZhao:
Fixes #1302

### DIFF
--- a/pages/Decorators.md
+++ b/pages/Decorators.md
@@ -467,11 +467,11 @@ class Line {
 function validate<T>(target: any, propertyKey: string, descriptor: TypedPropertyDescriptor<T>) {
     let set = descriptor.set;
     descriptor.set = function (value: T) {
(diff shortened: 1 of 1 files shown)


orta:
Thanks, we re-wrote the mixins page for the v2 site

PR #1314, opened 2020-04-09
Fix missing back-quote near "private modifier" in pages/release notes/TypeScript 3.8.md

karol-depka:
Fix missing back-quote near "private modifier" in Handbook (3.8 release notes)


orta:
Thanks! I came to apply this to the v2 site, but it looks like it was already done

PR #1319, opened 2020-04-19
Fix broken link on a website

randomhash:
Fixes #
https://www.typescriptlang.org/docs/handbook/Symbols.md#symboliterator
broken link, fix by example of types link construction, [here](https://www.typescriptlang.org/docs/handbook/enums.html#union-enums-and-enum-member-types)


randomhash:
it was Symbols.md#symboliterator, and link was broken, but anyways )

PR #1320, opened 2020-04-22
Fix out of sync example continuation

vassudanagunta:
The running example is based on the `Car` interface and `taxi` instance, but in the edited paragraph refers to `person` and `Person` without either being defined anywhere. Sticking with `taxi` and `Car` makes more sense and I suspect someone changed the example and simply missed this sentence.


orta:
Great spot! Thanks a lot - I've reapplied it to the new site in https://github.com/microsoft/TypeScript-Website/pull/832

PR #1321, opened 2020-04-29
Interfaces vs. Type Aliases hover update

ivancuric:
I tried the original method from and it didn't work in any TS from 3.33 to 3.83.

Hovering the return type, however, produces the intended result.


orta:
Thanks, I've just re-structured this section in a way that doesn't need these changes after I saw the same issues in https://github.com/microsoft/TypeScript-Website/pull/832

PR #1322, opened 2020-05-06
Add readonly for React Props interface

xiaoxiangmoe:
see https://reactjs.org/docs/components-and-props.html#props-are-read-only

### DIFF
--- a/pages/tutorials/React & Webpack.md
+++ b/pages/tutorials/React & Webpack.md
@@ -115,7 +115,7 @@ First, create a file named `Hello.tsx` in `src/components` and write the followi
 ```ts
 import * as React from "react";
 
-export interface HelloProps { compiler: string; framework: string; }
+export interface HelloProps { readonly compiler: string; readonly framework: string; }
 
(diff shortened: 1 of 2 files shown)


orta:
Yeah, that seems like a good call 👍

sw-yx:
fwiw i think a year ago @ferdaber said that the readonly is going to be applied from the `@types/react` side - not sure if that actually happened

ferdaber:
For class components I believe they are already read only.

xiaoxiangmoe:
For function component, readonly is needed.

What should I do now?

ferdaber:
this PR is probably fine as-is, the updates on the React typings side can come later.

orta:
I'm not merging any PRs on this repo (content is frozen in preparation for archiving - see #1313) but when we flip the switch, I'll port over PRs to the v2 site

So for now, you can do nothing

orta:
These two tutorials now redirect to the typescript pages on the tool's side - so I can't reapply any of this to the v2 site. Thanks for the PR though!

PR #1331, opened 2020-06-16
Fix unclear wording and typos

### LINKED ISSUE #1330: Typos in "Advanced Types"
```ts
pluck(taxi, ['year', 'unknown']); /
```

There is a random `/`.

```
(`object['42'`])
```

Backtick should be after the `]` so that the code is highlighted properly.

> As an example of some types that are immediately resolved, we can take a look at the following example

Saying the word "example" twice sounds awkward.

> But as an example of a place where conditional types are deferred - where they stick around instead of picking a branch - would be in the following

The grammar here is off enough that it also sounds awkward. Also, em dashes would be more appropriate here instead of keyboard's small dash.

johnjago:
This change fixes a couple small typos. It also makes a few sentences
more clear and less repetitive.

Fixes #1330

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -821,7 +821,7 @@ That means the compiler will check that you pass the right set of property names
 
 ```ts
 // error, 'unknown' is not in 'manufacturer' | 'model' | 'year'
-pluck(taxi, ['year', 'unknown']); /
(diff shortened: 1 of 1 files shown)


orta:
Thanks, it looks like most of these had already been fixed in the new handbook on the staging site!

PR #1334, opened 2020-07-12
Update React & Webpack.md

cseas:
Added a note to clarify the use of `externals` property in webpack config.

I followed this guide as my first webpack tutorial then moved on to webpack docs to learn more and add more things to the webpack config. None of the other tutorials I saw had manually added any React libraries to `index.html` so I removed those 2 lines and struggled for about a week to realize the `externals` field added from this tutorial was the reason my site stopped working.

I personally wouldn't recommend adding any `externals` to a beginner tutorial. Learning webpack is complex enough as it is. Something like `externals` should only be introduced in a "Production/Deployment" or "Optimizations" related page after a beginner has gone through the initial tutorial without hiccups. But if that has to be included in this beginner tutorial, the purpose of it should be made really clear.

### DIFF
--- a/pages/tutorials/React & Webpack.md
+++ b/pages/tutorials/React & Webpack.md
@@ -177,7 +177,7 @@ Create a file at the root of `proj` named `index.html` with the following conten
 </html>
 ```
 
-Notice that we're including files from within `node_modules`.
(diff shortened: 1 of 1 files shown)


orta:
Thanks, this page has been replaced with a redirect to the official webpack docs for TypeScript - so people shouldn't get caught up in it in the future

PR #1335, opened 2020-07-17
Fixed Tree data structure definition

### LINKED ISSUE #1339: Incorrect tree node definition in  pages/Advanced Types.md
This is minor but I'm reporting it to make my PR reference an issue.

The Tree type defines a node of a tree which must have two Tree chidren. This makes it impossible for there to be leaf nodes.

```ts
type Tree<T> = {
    value: T;
    left: Tree<T>;
    right: Tree<T>;
}
```
If a tree must have a child and that child can't be anything other than a tree, then that causes every child to need a child of its own. A leaf node doesn't have any children and you can't have an infinite tree, so this type wouldn't work in real code.

chanagaray:
The Tree type defines a node of a tree which must have two Tree chidren. This makes it impossible for there to be leaf nodes. This commit makes the Tree type definition realistic, making both subtrees optional.

Fixes #1339

### DIFF
--- a/pages/Advanced Types.md
+++ b/pages/Advanced Types.md
@@ -483,8 +483,8 @@ We can also have a type alias refer to itself in a property:
 ```ts
 type Tree<T> = {
     value: T;
-    left: Tree<T>;
-    right: Tree<T>;
+    left?: Tree<T>;
+    right?: Tree<T>;
 }
 ```
 


orta:
Thanks! Added this to the new site in https://github.com/microsoft/TypeScript-Website/pull/832

PR #1338, opened 2020-07-28
Add any type for overloading examples.

pkonieczniak:
I noticed that in two examples `x` argument in `pickCard` method has a missing type. 

I suggest adding any type for consistency.


orta:
Agreed! Thanks! I've ported this to the new site - https://github.com/microsoft/TypeScript-Website/pull/832

